PR #3, opened 2018-06-15
set up CI

vikram-narayan:
Make travis build work, and add `appveyor.yml` and other files to make the appveyor build work.


appveyor.yml
```diff
@@ -0,0 +1,85 @@
...
+    # See: http://stackoverflow.com/a/13751649/163740
+    CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\ci\\appveyor\\run_with_env.cmd"
+
+    # 1. Generated a token for appveyor at https://anaconda.org/quantopian/settings/access with scope api:write.
+    #   Can also be done via anaconda CLI with
+    #     $ anaconda auth --create --name my_appveyor_token
+    # 2. Generated secure env var below via appveyor's Encrypt data tool at https://ci.appveyor.com/tools/encrypt.
+    #   See https://www.appveyor.com/docs/build-configuration/#secure-variables.
+    ANACONDA_TOKEN:
+      secure: "ryxTeOgjDvz6Y25wr29GcUm1Y7iCOeNX96t1iXIp4bQzWCQszJ68+M6Zy6Kpg0C6"
```
richafrank: I'm not certain - Should we generate new tokens for this repo?

appveyor.yml
```diff
@@ -0,0 +1,85 @@
...
+
+    CONDA_ROOT_PYTHON_VERSION: "2.7"
+
+    PYTHON_ARCH: "64"
+    PANDAS_VERSION: "0.18.1"
+    NUMPY_VERSION: "1.11.1"
+    SCIPY_VERSION: "0.17.1"
```
richafrank: Does this library depend on scipy?

appveyor.yml
```diff
@@ -0,0 +1,85 @@
...
+  - powershell .\ci\appveyor\install.ps1
+  - SET PATH=%PYTHON%;%PYTHON%\Scripts;%PATH%
+
+  - conda info -a
+  - conda install conda=4.3.30 conda-build=3.0.28 anaconda-client=1.6.3 --yes -q
+  - conda list
+  # https://blog.ionelmc.ro/2014/12/21/compiling-python-extensions-on-windows/ for 64bit C compilation
+  - ps: copy .\ci\appveyor\vcvars64.bat "C:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\bin\amd64"
+
+  - conda create -n testenv --yes -q --use-local pip python=%PYTHON_VERSION% numpy=%NUMPY_VERSION% scipy=%SCIPY_VERSION% -c quantopian -c https://conda.anaconda.org/quantopian/label/ci
```
richafrank: `--use-local` Are we building other packages locally that this package depends on? Or is this future-proofing?

`-c https://conda.anaconda.org/quantopian/label/ci` Same question - are there ci packages on anaconda that we expect to install here, or is this future-proofing?

appveyor.yml
```diff
@@ -0,0 +1,85 @@
...
+  - activate testenv
+  - SET CACHE_DIR=%LOCALAPPDATA%\pip\Cache\pip_np%CONDA_NPY%py%CONDA_PY%
+  - python -m pip install -e .[tests] --cache-dir=%CACHE_DIR%
+  - python -m pip freeze | sort
+
+test_script:
+  - flake8 trading_calendars tests
+  - nosetests tests
```
richafrank: Do we have any doctests in the trading_calendars dir that should get tested here too, i.e. is filtering to just the tests dir useful? Same question for travis.

.travis.yml
```diff
@@ -22,10 +22,11 @@ before_install:
   - conda update -q conda
 
 install:
-  - conda create -n testenv --yes -c quantopian pip python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY_VERSION pandas=$PANDAS_VERSION scipy=$SCIPY_VERSION libgfortran=$LIBGFORTRAN_VERSION
+  - conda create -n testenv --yes -c quantopian pip python=$TRAVIS_PYTHON_VERSION pandas=$PANDAS_VERSION
   - source activate testenv
   - CACHE_DIR="$HOME/.cache/.pip/pip_np""$CONDA_NPY"
```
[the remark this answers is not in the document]
vikram-narayan: I think for now we just want to test with different versions of python/pandas, removed that

.travis.yml
```diff
@@ -4,9 +4,9 @@ sudo: false
 matrix:
   include:
     - python: 2.7
-      env: PANDAS_VERSION=0.18.1 NUMPY_VERSION=1.11.1 SCIPY_VERSION=0.17.1 LIBGFORTRAN_VERSION=3.0
+      env: PANDAS_VERSION=0.18.1
     - python: 3.5
-      env: PANDAS_VERSION=0.19.2 NUMPY_VERSION=1.12.1 SCIPY_VERSION=0.19.0 LIBGFORTRAN_VERSION=3.0
+      env: PANDAS_VERSION=0.19.2
```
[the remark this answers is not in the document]
vikram-narayan: updated this matrix to test python 2.5/3.5 with pandas 0.18.1/0.22.0

appveyor.yml
vikram-narayan: removed this value - realized we don't need it if zipline is handling conda builds, but if we end up having this repo handle its own conda builds, might be worth revisiting

appveyor.yml
vikram-narayan: removed both of those - had thought of it as future proofing, but I don't think we need it as of now

appveyor.yml
vikram-narayan: removed that stuff

appveyor.yml
vikram-narayan: we don't have any doctests so I think this should be okay for now

.travis.yml
```diff
@@ -4,9 +4,13 @@ sudo: false
...
     - python: 2.7
-      env: PANDAS_VERSION=0.18.1 NUMPY_VERSION=1.11.1 SCIPY_VERSION=0.17.1 LIBGFORTRAN_VERSION=3.0
+      env: PANDAS_VERSION=0.18.1
+    - python: 2.7
+      env: PANDAS_VERSION=0.22.0
+    - python: 3.5
+      env: PANDAS_VERSION=0.18.1
     - python: 3.5
-      env: PANDAS_VERSION=0.19.2 NUMPY_VERSION=1.12.1 SCIPY_VERSION=0.19.0 LIBGFORTRAN_VERSION=3.0
+      env: PANDAS_VERSION=0.22.0
```
richafrank: FWIW I think we could also spell this out as the product of the two vars:
```yaml
python:
  - 2.7
  - 3.5
env:
  - PANDAS_VERSION=0.18.1
  - PANDAS_VERSION=0.22.0
```

.travis.yml
```diff
@@ -22,10 +26,11 @@ before_install:
   - conda update -q conda
 
 install:
-  - conda create -n testenv --yes -c quantopian pip python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY_VERSION pandas=$PANDAS_VERSION scipy=$SCIPY_VERSION libgfortran=$LIBGFORTRAN_VERSION
+  - conda create -n testenv --yes -c quantopian pip python=$TRAVIS_PYTHON_VERSION pandas=$PANDAS_VERSION
```
richafrank: When we run tests on travis/appveyor, do we want to pin transitive dependencies, like numpy?

.travis.yml
```diff
@@ -22,10 +26,11 @@ before_install:
...
   - source activate testenv
-  - CACHE_DIR="$HOME/.cache/.pip/pip_np""$CONDA_NPY"
+  - CACHE_DIR="$HOME/.cache/.pip/pip_np"
```
richafrank: Any reason to specify the cache dir? Also, the `np` suffix for numpy is inaccurate now.

vikram-narayan: fixed

.travis.yml
vikram-narayan: I'd removed them for simplicity, but I suppose for test consistency it makes sense to pin them. Added these back, lmk if these dependency versions should be different

.travis.yml
vikram-narayan: that's much cleaner; fixed that

appveyor.yml
```diff
@@ -0,0 +1,87 @@
...
+  - powershell .\ci\appveyor\install.ps1
+  - SET PATH=%PYTHON%;%PYTHON%\Scripts;%PATH%
+
+  - conda info -a
+  - conda install conda=4.3.30 conda-build=3.0.28 anaconda-client=1.6.3 --yes -q
+  - conda list
+  # https://blog.ionelmc.ro/2014/12/21/compiling-python-extensions-on-windows/ for 64bit C compilation
+  - ps: copy .\ci\appveyor\vcvars64.bat "C:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\bin\amd64"
+
+  - conda create -n testenv --yes -q python=%PYTHON_VERSION% numpy=%NUMPY_VERSION% scipy=%SCIPY_VERSION% -c quantopian
```
richafrank: Do we depend on scipy? Does pandas have any other dependencies that need pinning? How about our direct dependencies of `toolz`, `lru-dict`, `flake8`, etc?

.travis.yml
```diff
@@ -22,10 +26,11 @@ before_install:
   - conda update -q conda
 
 install:
-  - conda create -n testenv --yes -c quantopian pip python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY_VERSION pandas=$PANDAS_VERSION scipy=$SCIPY_VERSION libgfortran=$LIBGFORTRAN_VERSION
+  - conda create -n testenv --yes -c quantopian pip python=$TRAVIS_PYTHON_VERSION numpy=$NUMPY_VERSION pandas=$PANDAS_VERSION scipy=$SCIPY_VERSION libgfortran=3.0
   - source activate testenv
-  - CACHE_DIR="$HOME/.cache/.pip/pip_np""$CONDA_NPY"
+  - CACHE_DIR="$HOME/.cache/.pip/pip_np"
```
richafrank: Where did we end up on a pip cache per numpy version? I think it was useful in zipline, and here we're still naming it `_np`, but seems like we're not using it for that purpose here.

appveyor.yml
vikram-narayan: removed scipy, added the test versions in setup.py. Should the other dependencies be in a requirements_test.txt?

.travis.yml
vikram-narayan: removed the cache dir

ssanderson: @vikram-narayan I'm not sure, but it seemed like @richafrank was asking if we wanted to have a separate cache dir per numpy version. Is there a reason we think we don't want any caching here if we had it for zipline?

vikram-narayan: earlier Rich had said that the cache dir is specified in zipline so that libs compiled against numpy's C API don't stomp on each other. That doesn't appear to be happening on travis, so I don't _think_ it's necessary here, though I'm not super knowledgeable about how dependencies work on travis/appveyor. fwiw empyrical's build matrix also has different versions of pandas/numpy and doesn't use separate cache dirs per numpy version

ssanderson: Gotcha. Yeah, I don't think we need separate cache dirs for different numpy builds, but I think we do want to cache pip downloads between builds so that we don't have to re-download them every time. It might be that travis already does that by default though?

ssanderson: From https://docs.travis-ci.com/user/caching/#pip-cache, it looks like we can just do `cache: pip` and that should handle our needs here?

vikram-narayan: fixed

appveyor.yml
ssanderson: We probably don't even need a separate `requirements_test.txt`. Just `requirements.txt` seems fine for this, since the only case where you're going to want exactly-pinned versions is going to be for testing/local dev.

vikram-narayan: sounds good, fixed

vikram-narayan:
talked over in person, put `-e .` in `requirements.txt` and removed the extra requires in `setup.py`, and just installed requirements.txt in travis/appveyor yaml files

PR #8, opened 2018-08-05
ENH: Add a 24/7 calendar.

ssanderson:
This PR adds a simple "always open" calendar.

I wanted something like this for testing purposes in Zipline, and it also seems like it'd be generally useful for markets like forex that are always open.


trading_calendars/always_open.py
```diff
@@ -0,0 +1,27 @@
...
+
+from trading_calendars import TradingCalendar
+
+
+class AlwaysOpenCalendar(TradingCalendar):
+    """A TradingCalendar for an exchange that's open every minute of every day.
+    """
+    @property
+    def name(self):
+        return '24-7'
```
vikram-narayan: for consistency with the other calendars, should the name be `AlwaysOpen`?

ssanderson: "24-7" felt to me like a closer analog to the existing names like "NYSE", "ICE", and "CME" than "AlwaysOpen". This is the name that someone would type when calling `get_calendar`, so it seems useful to have it be easy to type and remember. I would have just called this `24By7Calendar` or something like that, but you can't use numbers as leading characters in identifiers in Python.

vikram-narayan: ok, that makes sense. I wasn't sure if you wanted it to be accessible via `get_calendar` since it's a calendar mainly used for testing purposes, but I think it'll need to be added to the list of calendars in `calendar_utils.py` so that `get_calendar` knows about it

PR #9, opened 2018-09-05
ENH: add frankfurt exchange calendar

vikram-narayan:
add calendar for the Frankfurt stock exchange (Frankfurter Wertpapierbörse, FWB).

Note that there is a `TODO` regarding verifying the early close time (cc @jmccorriston).


vikram-narayan:
fixed so that Whit Monday is observed from 2010 onwards on the FWB calendar

jmccorriston:
LGTM

vikram-narayan:
thanks for review, will squash commits and merge

PR #11, opened 2018-09-14
Weekday Calendar

dmichalowicz:
Adds a 24/5 calendar.


tests/test_trading_calendar.py
```diff
@@ -575,9 +575,31 @@ def _get_session_block(self):
...
+                    raise AssertionError(
+                        f'Failed to find a suitable block of test sessions '
+                        f'for the {self.calendar.name} calendar. Does your '
+                        f'test include at least 5 contiguous sessions?'
+                    )
+                if (block_start - one_day == pre_session) and \
+                   (block_end + one_day == post_session):
+                    found_enclosed_block = True
+
+            return all_sessions[i:i + 3]
```
dmichalowicz: This whole change feels a bit hacky, so I'm open to better ideas to fix using "random" dates here.

tests/test_trading_calendar.py
```diff
@@ -575,9 +575,31 @@ def _get_session_block(self):
...
+            while not found_enclosed_block:
+                i += 1
+                try:
+                    block_start = all_sessions[i]
+                    block_end = all_sessions[i + 2]
+                    pre_session = all_sessions[i - 1]
+                    post_session = all_sessions[i + 3]
+                except IndexError:
+                    raise AssertionError(
+                        f'Failed to find a suitable block of test sessions '
```
ssanderson: We're definitely not going to be able to support f-strings in this repo any time soon.

trading_calendars/weekday_calendar.py
```diff
@@ -0,0 +1,25 @@
...
+    def name(self):
+        return '24/5'
+
+    @property
+    def tz(self):
+        return 'UTC'
+
+    @property
+    def open_time(self):
+        return time(0)
```
ssanderson: Hrm. The convention we have elsewhere in this repo is that the open time is the label for the first minute of trading in a calendar, which means that, e.g. the first minute of the NYSE calendar is 9:31, not 9:30, and the last minute is 4:00. It seems like the analogous thing here would be to have (0, 1) and (24, 0), but (24, 0) isn't valid... I'm not sure how much that matters in practice for these 24 hour calendars though.

tests/test_trading_calendar.py
ssanderson: Yeah, I'm not in love with this either. One option might be to just make this a public method to get an "interesting" range of three contiguous dates, and allow tests that need to override it to do so.

trading_calendars/weekday_calendar.py
dmichalowicz: I see. FWIW the always_open calendar uses this pattern. I could look into finding out how to possibly change them both if that's worth it?

tests/test_trading_calendar.py
dmichalowicz: I considered that, my only qualm was that it defining that method still relied on knowing the "implementation details" so-to-speak of `test_minutes_in_range`, but I'm probably just looking too far into it for a test. Making this a method would probably still be an improvement.

ssanderson: I think if you added that method and documented that it should return a range of three contiguous "interesting" dates, that would be sufficient encapsulation.

trading_calendars/weekday_calendar.py
ssanderson: I don't see a good way to fix this right now. I'd leave it for the time being.

dmichalowicz:
@ssanderson Does this look good to you?

tests/test_trading_calendar.py
```diff
@@ -570,10 +570,12 @@ def test_sessions_in_range(self):
                                             second_session_label)
         )
 
-    def _get_session_block(self):
-        # Try to find and return a (full session, early close session, full
-        # session) block.
-
+    def get_session_block(self):
```
ssanderson: It might be helpful to have this raise an error telling the caller to override it if the random sessions aren't contiguous. Otherwise a new calendar implementor will have to trace back a confusing failure to this.

dmichalowicz: This function will always return 3 sessions in a row; I suppose they may not be contiguous in the sense that a weekend could be in between, but as far as I can tell that shouldn't actually matter. So maybe I should just change the wording of the doc string to say "3 in a row" or "continuous".

PR #12, opened 2018-09-19
ENH: add Swiss calendar (XSWX)

vikram-narayan:


tests/test_six_calendar.py
```diff
@@ -0,0 +1,48 @@
+from unittest import TestCase
+import pandas as pd
+
+from .test_trading_calendar import ExchangeCalendarTestBase
+from trading_calendars.exchange_calendar_six import SIXExchangeCalendar
+
+
+class LSECalendarTestCase(ExchangeCalendarTestBase, TestCase):
```
richafrank: Noticed a copy/paste error - should be `SIXCalendarTestCase`

vikram-narayan: sorry, fixed

trading_calendars/exchange_calendar_six.py
```diff
@@ -0,0 +1,149 @@
...
+)
+
+AscensionDay = Holiday(
+    "Ascension Day",
+    month=1,
+    day=1,
+    offset=[Easter(), Day(39)],
+)
+
+WhitMonday = Holiday(
```
ssanderson: I notice this is duplicated with the definition in the German calendar, except that in germany they didn't observe this until 2010. Maybe we should add a `common_holidays.py` or something and put a helper function there like:

```
def whit_monday(observance):
    return Holiday("Whit Monday", ..., observance=observance)
```

cc @dmichalowicz this might also make sense for the calendars you're looking at.

trading_calendars/exchange_calendar_six.py
```diff
@@ -0,0 +1,149 @@
...
+    offset=[Easter(), Day(50)],
+)
+
+SwissNationalDay = Holiday(
+    "Swiss National Day",
+    month=8,
+    day=1
+)
+
+ChristmasEve = Holiday(
```
ssanderson: These seem like good candidates for a common holidays file as well.

ssanderson: Seems like everything here but "Berchtold's Day" and "Swiss National Day" could probably be shared.

ssanderson: (I'd probably rename labour day to something like "European Labour Day" though if we move it to a common location.)

trading_calendars/exchange_calendar_six.py
vikram-narayan: Yup, I'd thought about using the FWB's Whit Monday, and didn't because of the start date issue. Will move this and the Christmas-related holidays to that file

trading_calendars/exchange_calendar_six.py
vikram-narayan: 👍 moved a bunch of stuff to common_holidays, and removed their definitions in the specific calendars

trading_calendars/common_holidays.py
```diff
@@ -0,0 +1,87 @@
+from pandas.tseries.holiday import Holiday, Easter
+from pandas.tseries.offsets import Day
+
+from .trading_calendar import MONDAY, TUESDAY
+
+
+def new_years_day(start_date=None,
```
ssanderson: API-wise it seems a bit confusing to have a mix of factory functions and holiday objects in this file. What do you think about just having a consistent interface of:

```
def snake_case_holiday_name(start_date, end_date, offset, observance):
    return Holiday(...)
```
?

vikram-narayan: forgot to push this friday, just pushed

PR #13, opened 2018-09-25
Rename calendars by their ISO-10383 codes

vikram-narayan:
Rename calendars to use their [ISO-10383 market identifier code](https://www.iso20022.org/10383/iso-10383-market-identifier-codes), specifically:
- New York: NYSE -> XNYS
- London Stock Exchange: LSE -> XLON
- Toronto Stock Exchange: TSX -> XTSE
- Tokyo Stock Exchange: TSE -> XTKS
- Frankfurt Stock Exchange: FWB -> XFRA
- Chicago Mercantile Exchange: CME -> CMES
- Intercontinental Exchange: ICE -> IEPA
- CBOE Futures Exchange: CFE -> XCBF
- Brazilian Mercantile and Futures Exchange: BMF -> BVMF

I also added a `NYSE` alias so that `get_calendar('NYSE')` will still work


trading_calendars/exchange_calendar_xnys.py
```diff
@@ -65,7 +65,7 @@ class NYSEExchangeCalendar(TradingCalendar):
     - Thanksgiving (fourth Thursday in November)
     - Christmas (observed on nearest weekday to December 25)
 
-    NOTE: The NYSE does not observe the following US Federal Holidays:
+    NOTE: The XNYS does not observe the following US Federal Holidays:
```
ssanderson: I'm not sure we need to update the lines in this docstring. I think most people will read this easier as NYSE than as XNYS.

trading_calendars/us_holidays.py
```diff
@@ -30,7 +30,7 @@ def july_5th_holiday_observance(datetime_index):
 
 
 # These have the same definition, but are used in different places because the
-# NYSE closed at 2:00 PM on Christmas Eve until 1993.
+# XNYS closed at 2:00 PM on Christmas Eve until 1993.
```
ssanderson: "the XNYS" is a bit of an odd phrasing. I think this probably makes sense to leave as NYSE.

trading_calendars/exchange_calendar_xnys.py
```diff
@@ -75,7 +75,7 @@ class NYSEExchangeCalendar(TradingCalendar):
     - Christmas Eve (except on Fridays, when the exchange is closed entirely)
     - Day After Thanksgiving (aka Black Friday, observed from 1992 onward)
 
-    NOTE: Until 1993, the standard early close time for the NYSE was 2:00 PM.
+    NOTE: Until 1993, the standard early close time for the XNYS was 2:00 PM.
```
ssanderson: Same note here.

trading_calendars/exchange_calendar_xnys.py
```diff
@@ -47,9 +47,9 @@
 # http://www.stevemorse.org/jcal/whendid.html
 
 
-class NYSEExchangeCalendar(TradingCalendar):
+class XNYSExchangeCalendar(TradingCalendar):
     """
-    Exchange calendar for NYSE
+    Exchange calendar for NYSE (XNYS).
```
ssanderson: Should we just write this out as "New York Stock Exchange" to match the other calendars?

trading_calendars/exchange_calendar_cmes.py
```diff
@@ -40,7 +40,7 @@
 )
 
 
-class CMEExchangeCalendar(TradingCalendar):
+class CMESExchangeCalendar(TradingCalendar):
     """
     Exchange calendar for CME
```
ssanderson: Should we update this with a (CMES)? We appear to have done that for the other calendars.

trading_calendars/calendar_utils.py
```diff
@@ -4,42 +4,43 @@
...
-    'NYSE': NYSEExchangeCalendar,
-    'CME': CMEExchangeCalendar,
-    'FWB': FWBExchangeCalendar,
-    'ICE': ICEExchangeCalendar,
-    'CFE': CFEExchangeCalendar,
-    'BMF': BMFExchangeCalendar,
-    'LSE': LSEExchangeCalendar,
-    'TSX': TSXExchangeCalendar,
-    'TSE': TSEExchangeCalendar,
+    'XNYS': XNYSExchangeCalendar,
```
ssanderson: For backwards compat, I think we should restore aliases for the calendars we've already released (I think that's NYSE, CME, ICE, CFE, BMF, LSE, and TSX). Luckily, all of these except NYSE are 3 characters, so there's no risk of collision moving forward. Grepping internally, I see usages of `get_calendar` with `CME`, `ICE`, and `CFE`, so I think we depend on at least those.

In the future, we may need to change the ICE alias, since XICE is the official code for the NASDAQ Iceland, which means we probably have conflicting vendor data with an exchange code of ICE.

PR #14, opened 2018-09-26
Calendars for Countries on Euronext Exchange

dmichalowicz:
Calendars for Belgium, France, Netherlands and Portugal. They all share the same commonalities of the Euronext exchange, but each still have their own idiosyncrasies.


trading_calendars/country_calendar_belgium.py
```diff
@@ -0,0 +1,128 @@
...
+    whit_monday,
+    european_labour_day,
+    christmas_eve,
+    christmas,
+    boxing_day,
+)
+from .trading_calendar import HolidayCalendar, TradingCalendar, WEEKDAYS
+
+NewYearsEveBefore2002 = new_years_eve(end_date='2002', days_of_week=WEEKDAYS)
+NewYearsEve = new_years_eve(start_date='2002', days_of_week=WEEKDAYS)
```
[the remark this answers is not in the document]
vikram-narayan: also, should these be after BoxingDay and before the calendar definition in this file so that the holidays are in the order they occur?

trading_calendars/country_calendar_france.py
```diff
@@ -0,0 +1,125 @@
...
+    EasterMonday,
+    whit_monday,
+    european_labour_day,
+    christmas_eve,
+    christmas,
+    boxing_day,
+)
+from .trading_calendar import HolidayCalendar, TradingCalendar, WEEKDAYS
+
+NewYearsEveBefore2002 = new_years_eve(end_date='2002', days_of_week=WEEKDAYS)
```
vikram-narayan: do you need to explicitly pass WEEKDAYS for all these holidays? If they fall on a weekend, nothing will happen, right?

trading_calendars/common_holidays.py
```diff
@@ -1,12 +1,13 @@
-from pandas.tseries.holiday import Holiday, Easter
+from pandas.tseries.holiday import Holiday, GoodFriday, Easter, EasterMonday  # noqa
```
vikram-narayan: to be consistent with the rest of this file, should we make factory functions for Good Friday and Easter Monday? Otherwise we can just import them from pandas.tseries.holiday where needed

tests/test_belgium_calendar.py
```diff
@@ -0,0 +1,123 @@
...
+            pd.Timestamp('2001-05-24', tz='UTC'),  # Ascension Day
+            pd.Timestamp('2002-05-20', tz='UTC'),  # Whit Monday
+            pd.Timestamp('2003-07-21', tz='UTC'),  # Independence Day
+            pd.Timestamp('2002-12-31', tz='UTC'),  # New Year's Eve
+        ]
+
+        for session_label in expected_sessions:
+            self.assertIn(session_label, all_sessions)
+
+    def test_start_end(self):
```
vikram-narayan: I moved this test into `ExchangeCalendarTestBase` in the branch I recently merged, so you should be able to remove this test (and the corresponding tests for the other countries) after rebasing

trading_calendars/calendar_utils.py
```diff
@@ -50,6 +61,16 @@
     'ICEUS': 'IEPA',
     'NYFE': 'IEPA',
     'CFE': 'XCBF',
+    'ENX': 'XPAR',  # Unsure about this
+    'ZZFR': 'XPAR',
+    'COAL': 'XPAR',
+    'LIF': 'XPAR',
+    'ZZBE': 'XBRU',
+    'ZZPT': 'XLIS',
```
vikram-narayan: where are these from? I don't see them grepping for Portugal in the ISO code csv

trading_calendars/calendar_utils.py
```diff
@@ -30,6 +35,12 @@
     'XSWX': XSWXExchangeCalendar,
     'XTSE': XTSEExchangeCalendar,
     'XTKS': XTKSExchangeCalendar,
+    # Country calendars.
```
vikram-narayan: what is the distinction between country and exchange calendars? the country calendars are also exchange calendars, aren't they?

tests/test_belgium_calendar.py
```diff
@@ -0,0 +1,123 @@
+from unittest import TestCase
+import pandas as pd
+
+from .test_trading_calendar import ExchangeCalendarTestBase
+from trading_calendars.country_calendar_belgium import BelgiumCalendar
+
+
+class BelgiumCalendarTestCase(ExchangeCalendarTestBase, TestCase):
```
vikram-narayan: I agree these could be consolidated - maybe have a `EuronextCalendarTestBase(ExchangeCalendarTestBase)` which contains test_normal_year and test_holidays_fall_on_weekend?

trading_calendars/calendar_utils.py
dmichalowicz: The distinction was that they are all on the same exchange (Euronext) but we needed different calendars for all of them. Furthermore, the intention going in is that these calendars will encompass _all_ the exchanges in each country, so it seemed to make sense to just call them by country name. That being said, having looked at the ISO page of exchanges, there's an argument to be had for distinguishing them as "Euronext Brussels", "Euronext Paris", etc. instead, so maybe we want to do that now? @ssanderson @jmccorriston

trading_calendars/common_holidays.py
dmichalowicz: I'll make factory functions for them. I'd like to have them imported in this file.

dmichalowicz: jk that seems extraneous. I'll just move them

trading_calendars/calendar_utils.py
vikram-narayan: I think I'd be in favor of calling them Euronext Brussels, etc., since we already do that sort of thing for other exchanges, like aliasing Nasdaq to XNYS

trading_calendars/country_calendar_france.py
dmichalowicz: Actually it looks to only affect the half days (New Years Eve/Christmas Eve). I'll remove the rest.

dmichalowicz:
@vikram-narayan I think I addressed all of your comments if you don't mind another look.

trading_calendars/country_calendar_france.py
vikram-narayan: gotcha - does it make weekends seem like half days if you don't pass WEEKDAYS?

PR #15, opened 2018-09-27
TST: move test_start_end to base test

vikram-narayan:


dmichalowicz:
Just had the one comment, otherwise looks good.

vikram-narayan:
ah yup, I think made the change and then rebased, so I missed that. will fix and merge

PR #18, opened 2018-10-01
XNZE (New Zealand Exchange/NZX)

yankees714:
This also includes a fix for our special open/close logic, which was making an assumption that all open and closes occur on the UTC day corresponding to the session's label (that's not the case for early closes in New Zealand, which occur at 23:45 UTC on the previous day).


trading_calendars/exchange_calendar_xnze.py
```diff
@@ -0,0 +1,173 @@
...
+
+# Regular Holidays
+# ----------------
+NewYearsDay = new_years_day()
+
+DayAfterNewYearsDay = Holiday(
+    "Day after New Year's Day",
+    month=1,
+    day=2,
+)
```
dmichalowicz: Shouldn't this and `NewYearsDay` have a `weekend_to_monday` observance?

trading_calendars/exchange_calendar_xnze.py
```diff
@@ -0,0 +1,173 @@
...
+
+WeekendBoxingDay = weekend_boxing_day()
+
+
+# Early Closes
+# ------------
+BusinessDayPriorToChristmasDay = Holiday(
+    "Business Day prior to Christmas Day",
+    month=12,
+    day=25,
```
dmichalowicz: I think this should be `24`. This is basically like saying Christmas Eve is an early close unless it's a weekend, in which case go back to the previous workday.

trading_calendars/exchange_calendar_xnze.py
```diff
@@ -0,0 +1,173 @@
...
+    observance=previous_workday,
+    start_date="2011-01-01",
+)
+
+BusinessDayPriorToNewYearsDay = Holiday(
+    "Business Day prior to New Year's Day",
+    month=1,
+    day=1,
```
dmichalowicz: Same comment here, I believe this should be 12/31.

trading_calendars/exchange_calendar_xnze.py
dmichalowicz: Talked to Andrew in person, what he has actually works. I was thinking of `previous_friday`, which is confusingly similar:

```
> previous_workday?
Signature: previous_workday(dt)
Docstring: returns previous weekday used for observances

> previous_friday?
Signature: previous_friday(dt)
Docstring: If holiday falls on Saturday or Sunday, use previous Friday instead.
```

trading_calendars/trading_calendar.py
```diff
@@ -875,22 +884,57 @@ def minute_index_to_session_labels(self, index):
...
+            )
+            for time_, datetimes in ad_hoc_dates
+        ]
+
+        merged = regular + ad_hoc
+        if not merged:
+            # Concat barfs if the input has length 0.
+            return pd.Series([])
+
+        result = pd.concat(merged).sort_index()
```
dmichalowicz: 👍 on adding this sort.

PR #21, opened 2018-10-03
Austria Calendar

dmichalowicz:
Exchange calendar for the Wiener Borse AG exchange in Vienna.


trading_calendars/common_holidays.py
```diff
@@ -60,6 +75,62 @@ def whit_monday(start_date=None, end_date=None):
     )
 
 
+def corpus_christi(start_date=None, end_date=None):
```
yankees714: Looks like we also have these defined in some of the others calendars, should we update those to use these new ones?

trading_calendars/common_holidays.py
```diff
@@ -35,9 +50,9 @@ def european_labour_day(start_date=None,
     )
 
 
-# Whit Monday and Ascension day don't take observance as a parameter because
-# they depend on a particular offset, and offset and observance cannot both
-# be passed to a Holiday
+# Ascension Day, Whit Monday and Corpus Christi do not take observance as a
```
yankees714: Can't help myself, can we use an oxford comma here?

trading_calendars/exchange_calendar_xwbo.py
```diff
@@ -0,0 +1,126 @@
...
+NationalHoliday = Holiday('National Holiday', month=10, day=26)
+
+AllSaintsDay = all_saints_day()
+
+ImmaculateConception = immaculate_conception()
+
+ChristmasEve = christmas_eve()
+Christmas = christmas()
+
+SaintStephensDay = Holiday('Saint Stephens Day', month=12, day=26)
```
yankees714: I think this should be `Stephen's`

yankees714: Also wondering if we could combine this with Boxing Day?

trading_calendars/common_holidays.py
dmichalowicz: Ya, I was planning on updating this after those merged actually.

yankees714: Cool that's what I figured (thought it would be good to have a comment here so we don't forget)

trading_calendars/exchange_calendar_xwbo.py
dmichalowicz: > I think this should be Stephen's

Yup I'll fix

> Also wondering if we could combine this with Boxing Day?

I thought of this too but am inclined to say no just because it's a simple enough holiday to just give it its own name. cc @jmccorriston

trading_calendars/common_holidays.py
dmichalowicz: Yup thanks 👍

dmichalowicz: @yankees714 I made these updates, mind double checking them for me?

PR #22, opened 2018-10-05
Assorted calendar fixes

yankees714:
A few issues uncovered after comparing the exchange calendars against a third party source.

(See commits for details).


trading_calendars/exchange_calendar_xfra.py
```diff
@@ -127,6 +128,16 @@ def regular_holidays(self):
...
         ])
 
+    @property
+    def adhoc_holidays(self):
+        return [
+            # Whit Monday observed in 2007, before it became regularly
+            # observed starting in 2015.
+            Timestamp('2007-05-28', tz='UTC'),
+            # Reformation Day was a German national holiday in 2017.
+            Timestamp('2017-10-31', tz='UTC'),
```
vikram-narayan: For consistency with other adhocs, could define these as `WhitMonday2007`, `ReformationDay500thAnniversary` (or something to that effect) below the regular holidays in this file, and reference them here?

tests/resources/xfra.csv
```diff
@@ -7541,3 +7544,23 @@
 2019-09-03 00:00:00+00:00,2019-09-03 07:01:00,2019-09-03 15:30:00
 2019-09-04 00:00:00+00:00,2019-09-04 07:01:00,2019-09-04 15:30:00
 2019-09-05 00:00:00+00:00,2019-09-05 07:01:00,2019-09-05 15:30:00
+2019-09-06 00:00:00+00:00,2019-09-06 07:01:00,2019-09-06 15:30:00
```
vikram-narayan: what is causing these 2019 additions to each calendar's csv?

vikram-narayan:
looks like there's also a flake8 error

tests/resources/xfra.csv
yankees714: I think the calendars go a year out, so when you regenerate this output it adds the days out to a year from now. I can remove these if we want.

trading_calendars/exchange_calendar_xfra.py
yankees714: Sure, I might also add `AdHoc` to the end of those names to differentiate them from the holiday calendars

PR #31, opened 2018-10-22
Ireland Calendar

dmichalowicz:
Calendar for the Irish Stock Exchange in Dublin.


README.md
```diff
@@ -22,6 +22,7 @@ A Python library of exchange calendars meant to be used with [Zipline](https://g
 | New Zealand Exchange            | XNZE     | New Zealand | 1.3           | https://www.nzx.com/                                    |
 | Wiener Borse                    | XWBO     | Austria     | 1.3           | https://www.wienerborse.at/en/                          |
 | Hong Kong Stock Exchange        | XHKG     | Hong Kong   | 1.3           | https://www.hkex.com.hk/?sc_lang=en                     |
+| Irish Stock Exchange            | XDUB     | Ireland     | 1.4*          | http://www.ise.ie/                                      |
```
dmichalowicz: @jmccorriston According to [this](https://en.wikipedia.org/wiki/Euronext_Dublin) the Irish Stock Exchange is now Euronext Dublin. Should I change the name in my PR here?

tests/test_xdub_calendar.py
```diff
@@ -0,0 +1,161 @@
...
+            pd.Timestamp('2018-08-06', tz='UTC'),  # August Bank Holiday
+            pd.Timestamp('2018-10-29', tz='UTC'),  # October Bank Holiday
+            pd.Timestamp('2018-12-25', tz='UTC'),  # Christmas Day
+            pd.Timestamp('2018-12-26', tz='UTC'),  # Boxing Day
+        ]
+
+        for session_label in expected_holidays:
+            self.assertNotIn(session_label, all_sessions)
+
+        # June Bank Holiday did not start until 2012.
```
vikram-narayan: are we sure this is the case? searching around, I can't find anything specific to the stock exchange, but it looks as though the June bank holiday was observed before that according to https://www.officeholidays.com/countries/ireland/2010.php and some other sources I was looking at. I'm not sure how accurate those are, but it might be good to double check this

dmichalowicz: Thanks for catching this, I'm going to remove this start date.

trading_calendars/exchange_calendar_xdub.py
```diff
@@ -0,0 +1,178 @@
...
+        return HolidayCalendar([
+            NewYearsDay,
+            StPatricksDay,
+            GoodFriday,
+            EasterMonday,
+            LabourDay,
+            MayBankHoliday,
+            JuneBankHoliday,
+            AugustBankHoliday,
+            OctoberBankHoliday,
```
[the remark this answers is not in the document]
dmichalowicz: Will do

tests/test_xdub_calendar.py
```diff
@@ -0,0 +1,161 @@
+from unittest import TestCase
+import pandas as pd
+
+from .test_trading_calendar import ExchangeCalendarTestBase
+from trading_calendars.exchange_calendar_xdub import XDUBExchangeCalendar
+
+
+class XDUBCalendarTestBase(ExchangeCalendarTestBase, TestCase):
```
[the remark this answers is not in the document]
dmichalowicz: d'oh

dmichalowicz:
Not sure of the problem here...

trading_calendars/exchange_calendar_xdub.py
```diff
@@ -0,0 +1,157 @@
...
+    day=24,
+    start_date='2010',
+    observance=previous_friday,
+)
+Christmas = christmas()
+WeekendChristmas = weekend_christmas()
+BoxingDay = boxing_day()
+WeekendBoxingDay = weekend_boxing_day()
+
+LastTradingDayOfCalendarYear = Holiday(
```
dmichalowicz: This and `LastTradingDayBeforeChristmas ` are based on http://www.ise.ie/Media/News-and-Events/2016/Irish-Stock-Exchange-Trading-Calendar-for-2017.html.

README.md
jmccorriston: Does Euronext Dublin have a separate iso code? If so, I think we should stick with the Irish Stock Exchange name. If not, we can change it to Euronext Dublin. I'm a little bit worried that FactSet might change the exchange name on their end. What would happen if such a case were to arise?

dmichalowicz: I actually don't think if FactSet changed an exchange name that it would have a consequential effect on us downstream. Euronext Dublin does not appear to have a separate iso code, but that being said I'm going to live this as-is for now and we can revisit it if necessary.

PR #32, opened 2018-10-23
ENH: Tweak/update the BVMF (Brazil) calendar and add tests.

jbredeche:


tests/test_bvmf_calendar.py
```diff
@@ -0,0 +1,72 @@
+from unittest import TestCase
+import pandas as pd
+
+from .test_trading_calendar import ExchangeCalendarTestBase
+from trading_calendars.exchange_calendar_bvmf import BVMFExchangeCalendar
+
+
+class BVMFCalendarTestCase(ExchangeCalendarTestBase, TestCase):
+
```
dmichalowicz: Should we add a test case for the special open day (`QuartaCinzas`)?

jbredeche: yes, will do. I didn't see that special open day.

trading_calendars/exchange_calendar_bvmf.py
```diff
@@ -165,7 +179,8 @@ class BVMFExchangeCalendar(TradingCalendar):
     - Proclamation of the Republic (November 15)
     - Day of Black Awareness (November 20 after 2004)
     - Christmas (December 24 and 25)
-    - Day before New Year's Eve (December 30 if NYE falls on a Saturday)
+    - Day before New Year's Eve (December 30 if NYE falls on a Saturday,
+        December 29 if NYE falls on a Monday)
```
yankees714:
```suggestion
        December 29 if NYE falls on a Sunday)
```

trading_calendars/exchange_calendar_bvmf.py
```diff
@@ -165,7 +179,8 @@ class BVMFExchangeCalendar(TradingCalendar):
...
     - New Year's Eve (December 31)
```
yankees714: As I mentioned with the holiday definitions, I think we could probably combine this with the above, the day before is an observance, not a separate holiday. Another way to phrase this would be "Business Day before New Year's Day".

tests/test_bvmf_calendar.py
```diff
@@ -0,0 +1,72 @@
...
+            pd.Timestamp('2017-05-01', tz='UTC'),  # Labor Day
+            pd.Timestamp('2017-06-15', tz='UTC'),  # Corpus Christi Day
+            pd.Timestamp('2017-09-07', tz='UTC'),  # Independence Day
+            pd.Timestamp('2017-10-12', tz='UTC'),  # Our Lady of Aparecida Day
+            pd.Timestamp('2017-11-02', tz='UTC'),  # All Souls Day
+            pd.Timestamp('2017-11-15', tz='UTC'),  # Proclamation of the
+                                                   #  Republic Day
+            pd.Timestamp('2017-11-20', tz='UTC'),  # Black Consciousness Day
+            pd.Timestamp('2017-12-25', tz='UTC'),  # Christmas Day
+            pd.Timestamp('2017-12-29', tz='UTC'),  # Day before New Years
```
yankees714: We're not checking  New Year's Day and Constitutional Revolution Day here because they fall on the weekend, is there another year where these fall better? Or maybe we could just find a year in which they're observed and check multiple years here?

tests/test_bvmf_calendar.py
```diff
@@ -0,0 +1,72 @@
...
+
+        # 2014: Jan 1 is Wednesday, so Tuesday 12/31 should be closed
+        self.assertFalse(
+            self.calendar.is_session(pd.Timestamp('2013-12-31', tz='UTC'))
+        )
+
+    def test_world_cup(self):
+        self.assertFalse(
+            self.calendar.is_session(pd.Timestamp('2014-06-12', tz='UTC'))
+        )
```
yankees714: Might be worth adding a test for the holidays that have start dates (Constitutionalist Revolution Day, Black Consciousness Day), to make sure it's not observed before the start date.

trading_calendars/exchange_calendar_bvmf.py
jbredeche: good idea, done

tests/test_bvmf_calendar.py
jbredeche: good idea, I added some other years.

tests/test_bvmf_calendar.py
jbredeche: good idea, done

trading_calendars/exchange_calendar_bvmf.py
```diff
@@ -133,21 +133,35 @@
...
+)
 
 
 class BVMFExchangeCalendar(TradingCalendar):
     """
     Exchange calendar for BM&F BOVESPA (BVMF).
 
     Open Time: 10:00 AM, Brazil/Sao Paulo
-    Close Time: 4:00 PM, Brazil/Sao Paulo
+    Close Time: 5:30 PM, Brazil/Sao Paulo
```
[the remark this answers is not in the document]
jbredeche: I think it was wrong.  Upon further research, it looks like the regular session closes at 5pm, so I'll change it.

jbredeche:
@dmichalowicz @yankees714 thanks for the feedback, just pushed an update.

PR #39, opened 2018-11-01
BUG: Add missing XBOM holidays.

jbredeche:
Thanks @pvssvikas for finding a missing holiday, and I found some other ones upon more inspection.


pvssvikas:
Thanks a lot for getting this fixed, yes I tested only on 2018 data so might have not figured out others.

pvssvikas:
My pleasure. feels good to see the quick response 👍

PR #41, opened 2018-11-21
ENH: Add Korea Exchange (XKRX)

freddiev4:
This PR adds the Korea Exchange (XKRX) and uses a trusted source for the holidays up until the end of 2019.

@dmichalowicz could you take a look at this when you have time? Opening Zipline / other PRs as well.


trading_calendars/exchange_calendar_xkrx.py
```diff
@@ -0,0 +1,546 @@
...
+])
+
+
+class XKRXExchangeCalendar(PrecomputedTradingCalendar):
+    """
+    Calendar for the Korea exchange, and the primary calendar for
+    the country of South Korea.
+
+    Open Time: 9:00 AM, KST (Korean Standard Time)
+    Close Time: 3:30 PM, KST (Korean Standard Time)
```
dmichalowicz: I might also just add a comment saying they observe standard time all year round (i.e. no daylight savings time).

tests/test_xkrx_calendar.py
```diff
@@ -0,0 +1,64 @@
...
+            T("2017-10-09"),
+            T("2017-12-25"),
+            T("2017-12-29"),
+        ]
+
+        for session_label in expected_holidays_2017:
+            self.assertNotIn(session_label, self.calendar.all_sessions)
+
+    def test_constrain_construction_dates(self):
+        # the XKRX calendar currently goes from 1999 to 2019, inclusive.
```
dmichalowicz: `1999` --> `1986`?

trading_calendars/exchange_calendar_xkrx.py
```diff
@@ -0,0 +1,546 @@
...
+
+    Due to the complexity around the Korean holidays, we are hardcoding
+    a list of holidays covering 1986-2019, inclusive.
+    """
```
dmichalowicz: I don't know if it's worth it, but maybe you could add the list of Korean holidays in this docstring, like is done in other calendars? Just the names of them.

tests/test_xkrx_calendar.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class XKRXCalendarTestCase(NoDSTExchangeCalendarTestBase, TestCase):
+
+    answer_key_filename = 'xkrx'
+    calendar_class = XKRXExchangeCalendar
+
+    # Korea exchange is open from 9am to 3:30pm
+    MAX_SESSION_HOURS = 6.5
+    HAVE_EARLY_CLOSES = False
```
dmichalowicz: I see there are no early closes, but are there any holidays that may or may not fall on a weekend? In which case we would need to test that those holidays are either not observed some years, or properly roll to the next weekday? I'm just trying to think of any edge cases we might want to specifically test here, other than just a "normal" year.

freddiev4: I think there are at least two holidays that I can think of that _can_ occur on a weekend.

Is there anywhere we already have a test like that?

dmichalowicz: Here is an example of holidays falling on weekends that _are_ made up: https://github.com/quantopian/trading_calendars/blob/master/tests/test_xasx_calendar.py#L42

Here is an example of holidays falling on a weekend that _are not_ made up:https://github.com/quantopian/trading_calendars/blob/master/tests/test_xhel_calendar.py#L45

freddiev4:
@dmichalowicz updated the PRs with your feedback

trading_calendars/exchange_calendar_xkrx.py
```diff
@@ -521,8 +521,27 @@ class XKRXExchangeCalendar(PrecomputedTradingCalendar):
...
+    - Buddha's Birthday
+    - Memorial Day
+    - Provincial Election Day
+    - Liberation Day
+    - Chuseok (Korean Thanksgiving)
+    - National Foundation Day
+    - Christmas Day
+    - End of Year Holiday
+
+    NOTE: Hangeul Day became a national holiday in 2013
```
dmichalowicz: Could you add a short test asserting that Hangeul Day (October 9th) is a session in 2012 but not a session in 2013?

tests/test_xkrx_calendar.py
```diff
@@ -62,3 +62,36 @@ def test_constrain_construction_dates(self):
...
+            # Memorial Day on Sunday
+            T('2010-06-06'),
+        ]
+
+        for session_label in expected_holidays:
+            self.assertNotIn(session_label, self.calendar.all_sessions)
+
+        # Holidays below falling on a weekend should
+        # be made up during the week.
+        expected_sessions = [
```
dmichalowicz: I think you meant to call this `expected_holidays`?

tests/test_xkrx_calendar.py
```diff
@@ -62,3 +62,36 @@ def test_constrain_construction_dates(self):
                 ' cannot instantiate the XKRX calendar for 2020.'
             )
         )
+
+    def test_holidays_fall_on_weekend(self):
+        # Holidays below falling on a weekend should
+        # not be made up during the week.
+        expected_holidays = [
+            # National Foundation Day on a Saturday
+            T('2015-10-03'),
```
dmichalowicz: Since this is a Saturday (which is always not a session), I would actually instead test that the surrounding Friday and Monday are sessions (i.e. *not* holidays).

trading_calendars/exchange_calendar_xkrx.py
freddiev4: Sure! fwiw I think it should be the other way around; Hangeul Day should _not_ be a session in 2012, and only starts to be a Market-closing Holiday in 2013

freddiev4: nvm chatted IRL

PR #44, opened 2019-01-08
 Added XICE (Nasdaq Nordic Iceland)

oyvindwe:


tests/test_xice_calendar.py
```diff
@@ -0,0 +1,69 @@
...
+
+        expected_holidays = [
+            pd.Timestamp('2018-01-01', tz='UTC'),  # New Year's Day
+            pd.Timestamp('2018-03-29', tz='UTC'),  # Maundy Thursday
+            pd.Timestamp('2018-03-30', tz='UTC'),  # Good Friday
+            pd.Timestamp('2018-04-02', tz='UTC'),  # Easter Monday
+            pd.Timestamp('2018-04-19', tz='UTC'),  # First Day of Summer
+            pd.Timestamp('2018-05-01', tz='UTC'),  # Labour Day
+            pd.Timestamp('2018-05-10', tz='UTC'),  # Ascension Day
+            pd.Timestamp('2019-06-17', tz='UTC'),  # National Day
```
jbredeche: Did you mean 2018 here?

jbredeche: also I think you are missing Whit Monday (2018-05-21).  Can you add that?

jbredeche:
Thanks @oyvindwe, looks really good.  Two minor comments in the test code.

tests/test_xice_calendar.py
oyvindwe: National Day is tested for 2019, as it fell on a Sunday on 2018. I changed it to 2018 to `test_holidays_fall_on_weekend` so it's clearer.

I forgot Whit Monday - good catch (due to basing the test on XSTO). It's a holiday in the CSV test resource though.

oyvindwe:
Thanks for the feedback!

Commits have been squashed. You can do that when merging the PR though. That avoids force pushing commits.

tests/test_xice_calendar.py
jbredeche: got it re: National Day, makes sense, thanks.

jbredeche:
thanks @oyvindwe !  We'll cut a new release soon.

Issue #45, opened 2019-01-31
"TypeError: data is already tz-aware UTC" in get_calendar()


acowlikeobject:
```
from trading_calendars import get_calendar
get_calendar('CME')
```
Raises `TypeError: data is already tz-aware UTC, unable to set specified tz: UTC`.

This is a result of [this issue](https://github.com/pandas-dev/pandas/issues/23959) in pandas where `'UTC'` and `pytz.UTC` are not considered equal (I'm seeing this in `pandas` version `0.24.0`).

I believe the fix would be to change the return in `trading_calendars.trading_calendar.scheduled_special_times()`:
```
return pd.Series(
        index=pd.DatetimeIndex(days, tz='UTC'),
        data=days_at_time(days, time, tz=tz),
)
```
to:
```
return pd.Series(
        index=pd.DatetimeIndex(days, tz=pytz.UTC),
        data=days_at_time(days, time, tz=tz),
)
```
(Basically, replace `'UTC'` with `pytz.UTC` when creating the index.)

acowlikeobject:
For now, monkey patching works:

```
import pytz
import trading_calendars as tc

def scheduled_special_times(calendar, start, end, time, tz):
    days = calendar.holidays(start, end)
    return pd.Series(
        index=pd.DatetimeIndex(days, tz=pytz.UTC),
        data=tc.trading_calendar.days_at_time(days, time, tz=tz),
)
tc.trading_calendar.scheduled_special_times = scheduled_special_times

from trading_calendars import get_calendar
get_calendar('CME')
```

PR #48, opened 2019-02-14
Uniform use of `pytz` timezones.

Froskekongen:
This PR makes sure that all instances of the `UTC` timezone are `pytz` timezones. This fixes https://github.com/quantopian/trading_calendars/issues/45


trading_calendars/trading_calendar.py
```diff
@@ -41,9 +42,9 @@
 from .utils.preprocess import preprocess, coerce
 
 
-start_default = pd.Timestamp('1990-01-01', tz='UTC')
-end_base = pd.Timestamp('today', tz='UTC')
-# Give an aggressive buffer for logic that needs to use the next trading
+start_default = pd.Timestamp('1990-01-01', tz=UTC)
+end_base = pd.Timestamp('today', tz=UTC)
+# Give an aggressive b tuffer for logic that needs to use the next trading
```
richafrank: typo

.travis.yml
```diff
@@ -4,12 +4,12 @@ sudo: false
 
 python:
   - 2.7
```
richafrank: We still want to tests 2.7 with NUMPY_VERSION=1.11.3 PANDAS_VERSION=0.18.1. (Taken from [zipline](https://github.com/quantopian/zipline/blob/013cd1b0ef916024a7233301a2c127111e889434/.travis.yml#L47))

.travis.yml
```diff
@@ -4,12 +4,12 @@ sudo: false
...
-  - 3.5
```
richafrank: We still want to test 3.5 with NUMPY_VERSION=1.11.3 PANDAS_VERSION=0.18.1.

.travis.yml
Froskekongen: The travis CI build works, but the appveyor build is bugged. I can change the travis build, but the appveyor build won't work until the underlying bug gets fixed.

Also, I would suggest updating zipline to require newer versions of both pandas and numpy.

richafrank: > the appveyor build is bugged

Oh, is it a particular job in the matrix that's broken?

> Also, I would suggest updating zipline to require newer versions of both pandas and numpy.

Agreed - PRs welcome 😄 .

Froskekongen: In appveyor, the pandas version is not constrained correctly from the specification, so it tries using pandas 0.24.1 with numpy 1.11.3, which are incompatible.

Froskekongen:
@richafrank: There was no explicit need for testing with pandas 0.19. The problem is in the appveyor build, which I outlined in the other conversation.

.travis.yml
richafrank: Ah, looks like the `conda create -n testenv` step is missing `pandas=%PANDAS_VERSION%`

Froskekongen: Thanks, and fixed (:

richafrank:
Thanks @Froskekongen !

PR #52, opened 2019-04-18
add holiday 20181231 to XSHG holidays

txu2014:
this notice doesn't show in SSE English version. 
 This holiday is in 2019 Chinese version. 
 http://www.sse.com.cn/disclosure/announcement/general/c/c_20181220_4696473.shtml


jbredeche:
Thanks @txu2014 . We've verified that 2018-12-31 was indeed a holiday for XSHG.

PR #57, opened 2019-07-05
July 3rd 2019 is an early close, July 5th 2019 is not.

gmanoim-quantopian:


trading_calendars/us_holidays.py
```diff
@@ -28,7 +28,7 @@
 
 
 def july_5th_holiday_observance(datetime_index):
-    return datetime_index[datetime_index.year != 2013]
+    return datetime_index[datetime_index.year not in (2013, 2019)]
```
richafrank: Hmm, this errors for me - does it work for you? Perhaps we need something like `numpy.isin` but which supports numpy>=1.11.

trading_calendars/exchange_calendar_xnys.py
```diff
@@ -160,6 +160,7 @@ def special_closes_adhoc(self):
                         '1999-12-31',
                         '2003-12-26',
                         '2013-07-03',
+                        '2019-07-03'
```
richafrank: Purely stylistic, but want to add a trailing comma, so we avoid extra noise in subsequent PRs?

trading_calendars/us_holidays.py
gmanoim-quantopian: Yep - that was the bug. Fixed in https://github.com/quantopian/trading_calendars/pull/57/commits/32ef5009ae1af178df61b9d4c95101d45cb65020

trading_calendars/exchange_calendar_xnys.py
gmanoim-quantopian: Fixed

trading_calendars/us_holidays.py
gmanoim-quantopian: Ah I see what you meant - there's still an issue in 1.11.3

gmanoim-quantopian: Trying again in https://github.com/quantopian/trading_calendars/pull/57/commits/951a91150e7e7fd1a7cfc509723dd4a61531bd93

trading_calendars/us_holidays.py
```diff
@@ -92,10 +94,11 @@ def july_5th_holiday_observance(datetime_index):
     days_of_week=(MONDAY, TUESDAY, THURSDAY),
     start_date=Timestamp("1995-01-01"),
 )
-FridayAfterIndependenceDayExcept2013 = Holiday(
+FridayAfterIndependenceDayExcept2013and2019 = Holiday(
```
ehebert: Not sure if the policy has changed, and this should be `FridayAfterIndependenceDay2013AndAfter`.

gmanoim-quantopian: I don't know if this is true in the future, it only seems published to 2021: https://www.nyse.com/markets/hours-calendars

Issue #58, opened 2019-07-07
MAINT: Fix tests on Travis


gmanoim-quantopian:
Builds with 
Python: 3.5, PANDAS_VERSION=0.18.1 NUMPY_VERSION=1.11.3
fail on travis (linux 64) because conda is unable to resolve dependencies. 

https://travis-ci.org/quantopian/trading_calendars/jobs/554823104

willianpaixao:
@richafrank I guess this issue can be closed.

richafrank:
Ah, thanks!

PR #59, opened 2019-07-08
Conda install that works

richafrank:
Pin to working version of conda


richafrank:
For https://github.com/quantopian/trading_calendars/issues/58

PR #60, opened 2019-08-02
added NSE India calendar XNSE

suryafx:
Added National Stock Exchange, India XNSE calendar with Holidays matched up to the year 2000


trading_calendars/exchange_calendar_xnse.py
```diff
@@ -0,0 +1,366 @@
...
+    '2019-09-02',
+    '2019-09-10',
+    '2019-10-02',
+    '2019-10-08',
+    '2019-10-27',
+    '2019-10-28',
+    '2019-11-12',
+    '2019-12-25',
+    '2020-02-21',
+    '2020-03-09',
```
jmccorriston: Should this be `2020-03-10` instead?

trading_calendars/exchange_calendar_xnse.py
```diff
@@ -0,0 +1,366 @@
...
+    '2020-04-02',
+    '2020-04-06',
+    '2020-04-10',
+    '2020-04-14',
+    '2020-04-1',
```
jmccorriston: This one looks like a typo, should it be removed?

trading_calendars/exchange_calendar_xnse.py
```diff
@@ -0,0 +1,366 @@
...
+    '2020-05-01',
+    '2020-07-31',
+    '2020-08-20',
```
jmccorriston: I don't see this one on the NSE website's list, what was the reason for the closure?

trading_calendars/exchange_calendar_xnse.py
```diff
@@ -0,0 +1,366 @@
...
+    '2020-10-02',
```
jmccorriston: I think there are two dates missing here: `2020-11-16` and `2020-11-30`.

jmccorriston:
Actually, do you know if this calendar is expected to differ from the Bombay Stock Exchange calendar?

theLufenk:
@jmccorriston 
AFAIK, BSE and NSE do not differ in the terms of trading days or open/close times.

There is another problem though. How to factor in extra working days. For instance, Saturday Feb 1st, 2020 was supposed to be a holiday, but the market was open since the Union Budget was presented on that day.

I tried overriding the `special_opens_adhoc` property, but apparently it's purpose is to providing a special opening time for days which are already in the calendar. How do we go about extra working days that fall within the Holiday rules ?

jmccorriston:
Hi @theLufenk - Currently, we don't have support for Saturday/Sunday work days. This is actually the main reason why we haven't added calendars for Israel, Egypt, Qatar, and a couple of other markets. The Moscow calendar is another example that has weekend days where the market is open, but those are not implemented in the current implementation. Supporting ad-hoc weekend work days is something we'd like to add in the long term, but unfortunately it's not in our short term plans. I'm not sure yet how we'd implement it exactly.

jmccorriston:
I'm going to close this PR for now since I think a more appropriate change would be to make NSE an alias of the BOM calendar.

PR #66, opened 2019-08-24
add holiday 2019-05-02 2019-05-03 in XSHG 

hustwd:
add holidays in XSHG


dmichalowicz:
Hi @hustwd, thanks for the PR! It looks like there are some test failures from this change, but I think they should be fixed by removing these two days from the [test CSV file](https://github.com/quantopian/trading_calendars/blob/master/tests/resources/xshg.csv#L4922-L4923).

PR #67, opened 2019-08-30
DOC: add XLON to table

llllllllll:
I believe `XLON` was added in 1.0, but just didn't make it into the table for some reason.


gmanoim-quantopian:
LGTM

PR #69, opened 2019-09-18
Chile Calendar

dmichalowicz:
Calendar for the Santiago Stock Exchange.


dmichalowicz:
@peterhbromley Thanks, I added a script in `etc` for generating the test csv files.

PR #72, opened 2019-09-20
Czech Republic Calendar

peterhbromley:


trading_calendars/exchange_calendar_xpra.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+NewYearsDay = new_years_day()
+
+GoodFriday = Holiday(
+    "Good Friday",
+    month=1,
+    day=1,
+    offset=[Easter(), Day(-2)],
+    start_date='2013'
```
dmichalowicz: This might be worth a quick comment that we can't use the built-in `GoodFriday` since we need a start date.

trading_calendars/exchange_calendar_xpra.py
```diff
@@ -0,0 +1,164 @@
...
+    Regularly-Observed Holidays:
+    - New Years Day
+    - Good Friday
+    - Easter Monday
+    - Labour Day (May 1)
+    - Liberation Day (May 8)
+    - Saints Cyril and Methodius Day (Jul 5)
+    - Jan Hus Day (Jul 6)
+    - Czech Statehood Day (Sep 28)
+    - Independence Day (Oct 28)
```
dmichalowicz: Missing `StruggleForFreedomAndDemocracyDay` here.

peterhbromley: Good catch!

trading_calendars/exchange_calendar_xpra.py
peterhbromley: Agreed, will do

PR #74, opened 2019-09-24
Fix and Test Poland Calendar

dmichalowicz:
In the XWAR `regular_holidays` method we were not actually invoking most of the holiday helper functions.

This PR also includes a fix for the XLIM open time.


trading_calendars/exchange_calendar_xwar.py
```diff
@@ -12,92 +14,139 @@
...
+    christmas,
+    boxing_day,
+    new_years_eve,
 )
+from .trading_calendar import HolidayCalendar, TradingCalendar
+
+
+def not_2004(datetime_index):
+    """
+    Christmas Eve is a holiday every year except for whatever reason it was a
```
gmanoim-quantopian: Fascinating.

trading_calendars/exchange_calendar_xwar.py
```diff
@@ -12,92 +14,139 @@
...
+    tz = timezone('Europe/Warsaw')
 
     open_times = (
-        (None, time(8, 30)),
+        (None, time(9, 1)),
     )
-
     close_times = (
-        (None, time(17, 5)),
+        (None, time(17)),
```
gmanoim-quantopian: gpw.pl/session-details has slightly different hours, not sure how we want to model that.

dmichalowicz: @jmccorriston Do you have thoughts on this? I'm going to merge for now, and if we decide to revisit this we'll need to also address some of the other calendars as well.

PR #75, opened 2019-09-24
Hungary Calendar

peterhbromley:


tests/test_xbud_calendar.py
```diff
@@ -0,0 +1,169 @@
...
+            self.assertIn(session_label, all_sessions)
+
+    def test_new_years_eve(self):
+        """
+        New Year's Eve became a holiday in 2011.  Before that it was not
+        a holiday unless it fell on a Monday (because of New Year's Day
+        four day weekend rule)
+        """
+        all_sessions = self.calendar.all_sessions
+        session_days = (1, 2, 3, 4)
```
dmichalowicz: In `trading_calendar.py` we define these days of weeks as constants (i.e. `TUESDAY=1`). That would be a little clearer to use here.

peterhbromley: Oh right, I must've forgotten about that.  And oops forgot to update the README, doing that now (and will include the 1.9s)!

PR #81, opened 2019-09-27
ENH: Add Johannesburg Stock Exchange (XJSE, South Africa)

quantophred:


tests/test_xjse_calendar.py
```diff
@@ -0,0 +1,61 @@
...
+
+    # The JSE is open from 09:00 to 17:00.
+    MAX_SESSION_HOURS = 8
+
+    HAVE_EARLY_CLOSES = False
+
+    # Johannesburg does not use daylight savings time: South Africa
+    # Standard Time (SAST) is observed all year.
+
+    def test_no_weekend_sessions(self):
```
peterhbromley: I like the way you did these tests, but I would maybe add a couple more.  Two tests that I can think of off the top of my head are: 1) testing that holidays that fall on Sundays get made up on Monday, but holidays that fall on Saturday do not, and 2) testing ad hoc holidays.

trading_calendars/exchange_calendar_xjse.py
```diff
@@ -0,0 +1,139 @@
...
+        (None, time(9, 1)),
+    )
+
+    close_times = (
+        (None, time(17, 00)),
+    )
+
+    @property
+    def regular_holidays(self):
+        return HolidayCalendar([
```
peterhbromley: We've been defining holidays as variables outside of the class definition, rather than in `regular_holidays()`.  I'm sure you saw this, so maybe you have a reason for defining them here.  I don't see anything wrong with defining them here, but maybe for uniformity it would be good to use the same structure as the rest of the calendars?  Just a thought.

trading_calendars/exchange_calendar_xjse.py
```diff
@@ -0,0 +1,139 @@
...
+)
+from pytz import timezone, UTC
+
+from .common_holidays import new_years_day
+from .trading_calendar import HolidayCalendar, TradingCalendar
+
+
+class XJSEExchangeCalendar(TradingCalendar):
+    """
+    Exchange calendar for the Johannesburg Stock Exchange (XJSE).
```
peterhbromley: In other calendars we've been formatting this docstring as follows:
```
    """
    Exchange calendar for the Istanbul Stock Exchange (XIST).

    Open Time: 9:30 AM, TRT
    Close Time: 5:30 PM, TRT

    Regularly-Observed Holidays:
    - New Year's Day
    - National Sovereignty and Children's Day (Apr 23)
    - Labour Day (May 1)
    - Commemoration of Attaturk, Youth and Sports Day (May 19)
    - Eid-ul-Fitr (Islamic Lunar Calendar)
    - Democracy and National Unity Day (Jul 15)
    - Eid-al-Adha (first day is a half-day, remaining full holidays)
    - Victory Day (Aug 30)
    - Republic Day (Oct 29)

    Early Closes:
    - First Day of Eid-al-Adha
    - Republic Day (Oct 28)
    """
```

tests/test_xjse_calendar.py
```diff
@@ -0,0 +1,61 @@
...
+    answer_key_filename = 'xjse'
+    calendar_class = XJSEExchangeCalendar
+
+    # The JSE is open from 09:00 to 17:00.
+    MAX_SESSION_HOURS = 8
+
+    HAVE_EARLY_CLOSES = False
+
+    # Johannesburg does not use daylight savings time: South Africa
+    # Standard Time (SAST) is observed all year.
```
peterhbromley: Looking at `ExchangeCalendarTestBase`, it looks like DAYLIGHT_SAVINGS_DATES defaults to `["2004-04-05", "2004-11-01"]`, so I don't think this is what you're looking for.  From quickly poking around other calendars that don't have daylight savings, I found that the Hong Kong calendar test case does this:
```
# Asia/Hong_Kong does not have daylight savings
    DAYLIGHT_SAVINGS_DATES = []
```

However, I'm not really sure what this test actually does because we don't define daylight savings anywhere else (maybe it's calculated from the given timezone, but then I'm not sure why this test passed when the default daylight savings dates were used here?)

quantophred: The test that uses these dates just checks the calculated open time for the next day's session: since this calendar's timezone doesn't apply daylight savings, the two open times it checks here are the same as all the others, so these checks are redundant.

Specifying this value as an empty list will essentially make the test a no-op. Neither of those possibilities seems super helpful, but maybe the latter will make it clearer that the test isn't doing anything useful.

trading_calendars/exchange_calendar_xjse.py
quantophred: I generally prefer to avoid additional name assignments when the object being constructed describes itself with a string literal, but it does look a little cluttered this way. I'll give it a shot.

tests/test_xjse_calendar.py
quantophred: Adding a test for uniform Sunday-to-Monday logic seems good.

Since the ad-hoc holidays are just a static declaration, I think they're covered by testing the full set of 2019 holidays: if one of the ad-hoc holidays is present there, we can pretty safely assume they're working correctly.

Iterating through all the ad-hoc holidays to verify their presence would basically just be a re-statement of the `adhoc_holidays` definition, which usually isn't a very useful kind of test, and makes it annoying to make changes; but something like adding another full year's worth of holidays could help, since it's a different formulation of the definition ("the set of all holidays in a specific year", rather than "the set of statically defined holidays across all years").

(`test_no_weekend_sessions` might actually not be very useful itself, in that it's basically a re-statement of logic defined elsewhere. I think I'll merge it with the Sunday-to-Monday test.)

PR #82, opened 2019-10-01
Turkey Calendar

peterhbromley:


trading_calendars/common_holidays.py
```diff
@@ -773,3 +773,159 @@ def chinese_national_day(start_date=None, end_date=None, observance=None):
...
+    '2041-12-04',
+    '2042-11-23',
+    '2043-11-12',
+    '2044-10-31',
+    '2045-10-21',
+    '2046-10-10',
+    '2047-09-30',
+    '2048-09-19',
+    '2049-09-08',
+])
```
dmichalowicz: Wow, interesting how it gradually changes date each year.

trading_calendars/exchange_calendar_xist.py
```diff
@@ -0,0 +1,208 @@
...
+    Early Closes:
+    - First Day of Eid-al-Adha
+    - Republic Day (Oct 28)
+    """
+    name = 'XIST'
+
+    tz = timezone('Europe/Istanbul')
+
+    open_times = (
+        (None, time(10, 1)),
```
dmichalowicz: It this time is "Effective as of Nov 14, 2016" does that mean it was something different before that? Is that something we need to encode?

trading_calendars/exchange_calendar_xist.py
```diff
@@ -0,0 +1,208 @@
...
+
+        return list(chain(
+            misc_adhoc,
+            EidAlFitr1,
+            EidAlFitr2,
+            EidAlFitr3,
+            EidAlAdha1,
+            EidAlAdha2,
+            EidAlAdha3,
+            EidAlAdha4,
```
dmichalowicz: Do we care that these could include weekends? I don't think so, but just double checking.

trading_calendars/exchange_calendar_xist.py
```diff
@@ -0,0 +1,208 @@
...
+            pd.Timestamp('1994-05-19'),
+            # Day before Eid al Fitr observed as holiday in 2003
+            pd.Timestamp('2003-11-24'),
+        ]
+
+        early_close_days = EidAlFitrHalfDay + EidAlAdhaHalfDay
+
+        early_close_days = [
+            day for day in early_close_days if day not in collisions
+        ]
```
dmichalowicz: What you have here is good, but just in case you find it useful another way to write this would be:

```python
collisions = set([
    # CAYS Day on May 19
    pd.Timestamp('1994-05-19'),
    # Day before Eid al Fitr observed as holiday in 2003
    pd.Timestamp('2003-11-24'),
])

early_close_days = list(set(EidAlFitrHalfDay + EidAlAdhaHalfDay) - collisions)
```

It's shorter, but yours is clearer I'd say.

tests/test_xist_calendar.py
```diff
@@ -0,0 +1,125 @@
...
+class XISTCalendarTestCase(ExchangeCalendarTestBase, TestCase):
+
+    answer_key_filename = 'xist'
+    calendar_class = XISTExchangeCalendar
+
+    # The XIST is open from 10:00 am to 6:00 pm
+    MAX_SESSION_HOURS = 8.0
+
+    # Daylight savings stopped being observed in 2017 in Turkey
+    DAYLIGHT_SAVINGS_DATES = []
```
dmichalowicz: In `test_trading_calendar.py` there's a `NoDSTExchangeCalendarTestBase` class, should we use that as the base class for this test suite?

tests/test_xist_calendar.py
```diff
@@ -0,0 +1,125 @@
...
+            pd.Timestamp('2003-11-21', tz=UTC),  # Terror attacks
+            pd.Timestamp('2004-01-23', tz=UTC),  # Bad weather
+            pd.Timestamp('2004-12-30', tz=UTC),  # Closure for redenomination
+            pd.Timestamp('2004-12-31', tz=UTC),  # Closure for redenomination
+            # Eid al Adha and Eid al Fitr extra closures
+            pd.Timestamp('2003-02-14', tz=UTC),  # Eid al Adha extra holiday
+            pd.Timestamp('2003-11-24', tz=UTC),  # Eid al Fitr extra holiday
+            pd.Timestamp('2003-11-28', tz=UTC),  # Eid al Fitr extra holiday
+            pd.Timestamp('2006-01-13', tz=UTC),  # Eid al Adha extra holiday
+        ]
```
dmichalowicz: 👍

trading_calendars/exchange_calendar_xist.py
peterhbromley: I talked with Jamie about this and because we don't actually know the open and close time before the effective date, we are just gonna leave it as 10am to 6pm for now.

trading_calendars/exchange_calendar_xist.py
peterhbromley: I don't think we do in this case, because these are defining full holidays so if they fall on a weekend the redundancy shouldn't matter.  We do care when defining early closures, because it doesn't make sense to say that there was an early closure on a holiday.

tests/test_xist_calendar.py
peterhbromley: Oh I didn't even realize we have this.  We could use that class, although looking at the daylight savings test case, making `DAYLIGHT_SAVINGS_DATES = []` bypasses the test as well.  However, it's probably cleaner to use the `NoDSTExchangeCalendarTestBase` so I'm going to make that change.

trading_calendars/exchange_calendar_xist.py
dmichalowicz: Sounds good

PR #83, opened 2019-10-02
Add Indonesia Stock Exchange (XIDX) 

quantophred:


trading_calendars/exchange_calendar_xidx.py
```diff
@@ -0,0 +1,447 @@
...
+        '2006-05-26',
+        '2007-05-18',
+
+        # Common leave for Independence Day.
+        '2003-08-18',
+        '2006-08-18',
+        '2008-08-18',
+
+        # Common leave for Christmas.
+        '2002-12-24',
```
peterhbromley: So there was no predictable pattern for when these extra Christmas days were observed?

tests/test_xidx_calendar.py
```diff
@@ -0,0 +1,151 @@
...
+            '2019': (245, [22, 19, 20, 19, 21, 15, 23, 22, 21, 23, 21, 19]),
+            '2020': (244, [22, 20, 21, 21, 14, 21, 22, 19, 22, 21, 21, 20]),
+        }
+        for year, (total_days, monthly_days) in expected_trading_days.items():
+            # Sanity checks for the test data.
+            self.assertEqual(len(monthly_days), 12, year)
+            self.assertEqual(sum(monthly_days), total_days, year)
+
+            sessions = self.calendar.sessions_in_range(year, year)
+            self.assertEqual(total_days, len(sessions), year)
```
peterhbromley: I like this test as a sanity check and verification against the actual stock exchange's website!

trading_calendars/exchange_calendar_xidx.py
quantophred: Not that I've been able to find, either from trading data or information online. Note that 2006, 2008, and 2015 had no additional leave.

PR #85, opened 2019-10-09
Russia Calendar

dmichalowicz:
Calendar for the Moscow Stock Exchange.


tests/test_xmos_calendar.py
```diff
@@ -0,0 +1,130 @@
...
+    # the XMOS calendar so overwrite that date here.
+    MINUTE_INDEX_TO_SESSION_LABELS_START = pd.Timestamp('2019-01-04', tz=UTC)
+    MINUTE_INDEX_TO_SESSION_LABELS_END = pd.Timestamp('2019-04-04', tz=UTC)
+
+    # None of the dates in the default range for these are trading days in the
+    # XMOS, so overwrite them here.
+    TEST_START_END_FIRST = pd.Timestamp('2019-01-13', tz=UTC)
+    TEST_START_END_LAST = pd.Timestamp('2019-01-20', tz=UTC)
+    TEST_START_END_EXPECTED_FIRST = pd.Timestamp('2019-01-14', tz=UTC)
+    TEST_START_END_EXPECTED_LAST = pd.Timestamp('2019-01-18', tz=UTC)
```
peterhbromley: I don't think I've been checking that these are valid dates for previous calendars... would there be a failed test if the default was not changed but fell on non-trading days?  If not, I should probably go back and make sure these are valid dates for the calendars I've done.

trading_calendars/exchange_calendar_xmos.py
```diff
@@ -0,0 +1,304 @@
...
+    pd.Timestamp('2007-06-11', tz=UTC),
+    pd.Timestamp('2008-06-13', tz=UTC),
+    pd.Timestamp('2012-06-11', tz=UTC),
+    pd.Timestamp('2014-06-13', tz=UTC),
+]
+
+unity_day_extensions = [
+    pd.Timestamp('2008-11-03', tz=UTC),
+    pd.Timestamp('2010-11-05', tz=UTC),
+]
```
peterhbromley: Are all of these "bridge holidays"?  And do the bridge holidays happen every year?  If so, you could also define a bridge holiday by, for example, defining a holiday one day after the real holiday that is only observed on Friday (with a custom `only_friday` observance function).

However, from a quick scan of these dates it looks like these bridge holidays might not be super consistent every year so maybe defining another Holiday wouldn't really work.

tests/test_xmos_calendar.py
dmichalowicz: > would there be a failed test if the default was not changed but fell on non-trading days?

Yes, one of [these checks](https://github.com/quantopian/trading_calendars/blob/master/tests/test_trading_calendar.py#L828-L835) would fail. Which means if all of your previous calendars are passing this test you shouldn't need to change anything.

In my case, I was actually hitting an error [on this line](https://github.com/quantopian/trading_calendars/blob/master/tests/test_trading_calendar.py#L826) during construction of the calendar, even before those assertions, because there were no trading days in the default range _at all_.

trading_calendars/exchange_calendar_xmos.py
dmichalowicz: > you could also define a bridge holiday by...

I did think of this (yes these are all bridge holidays), but from what I can tell these observances are not consistent every year (for example, Day of Russia is on a Tuesday in 2018 but the Monday before is not a holiday). So if I were to make a rule out of the bridge days I'd likely end up with just as many exceptions to that rule as adhoc days here, which would add yet another layer of confusion. My concern though is if that _is_ the rule in future years we will be missing coverage without manually revisiting this calendar every year. Either way you slice it though there's potential for incorrect coverage here, so I will add a comment to note all of this. @jmccorriston feel free to weigh in here, but for now I'm going to leave these as adhoc.

tests/test_xmos_calendar.py
peterhbromley: Oh okay, noted!

trading_calendars/exchange_calendar_xmos.py
peterhbromley: Yeah, in that case I think you're right that introducing more Holidays would just lead to more exceptions and we will have to revisit every year anyways, so ad-hoc seems like the way to go.

PR #87, opened 2019-10-09
Philippine Calendar

peterhbromley:


trading_calendars/exchange_calendar_xphs.py
```diff
@@ -0,0 +1,425 @@
...
+    month=8,
+    day=21,
+    start_date='2011',
+)
+
+NationalHeroesDay = Holiday(
+    'National Heroes Day',
+    month=8,
+    day=31,
+    observance=previous_monday,
```
dmichalowicz: I'm not positive if this works, but another way to do this might be to, instead of using `observance`, use `offset=DateOffset(weekday=MO(-1))`, where `MO` stands for Monday (`from dateutil.relativedelta import MO`)

trading_calendars/exchange_calendar_xphs.py
```diff
@@ -0,0 +1,425 @@
...
+
+
+# All pre-2011 holidays are pre-computed, so we define Holidays starting
+# in 2011.
+NewYearsDay = new_years_day(start_date='2011')
+
+# Chinese New Year's only observed after 2011
+ChineseNewYear = chinese_lunar_new_year_dates
+
+ChineseNewYearAfter2011 = ChineseNewYear[ChineseNewYear.year > 2011]
```
dmichalowicz: Did you mean `>= 2011`? Wondering because all of the other holidays start in 2011, but this is starting in 2012.

trading_calendars/exchange_calendar_xphs.py
peterhbromley: I forgot about DateOffset!  Made this change

trading_calendars/exchange_calendar_xphs.py
peterhbromley: I wondered about this too, but they didn't celebrate Chinese New Year in 2011 either.

trading_calendars/exchange_calendar_xphs.py
```diff
@@ -0,0 +1,425 @@
...
+philippines_eid_al_adha = pd.to_datetime([
+    '2011-11-07',
+    '2012-10-26',
+    '2013-10-15',
+    '2014-10-06',
+    '2015-09-25',
+    '2016-09-12',
+    '2017-09-01',
+    '2018-08-21',
+    '2019-08-11',
```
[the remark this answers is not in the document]
peterhbromley: I agree, it could get confusing and I think having one file with all of the hard-coded holidays would help.

PR #88, opened 2019-10-11
Thailand Calendar

dmichalowicz:
Calendar for the Stock Exchange of Thailand in Bangkok.


trading_calendars/xbkk_holidays.py
```diff
@@ -0,0 +1,319 @@
...
+    '2011-05-17',
+    '2012-06-04',
+    '2013-05-24',
+    '2014-05-13',
+    '2015-06-01',
+    '2016-05-20',
+    '2017-05-10',
+    '2018-05-29',
+    '2019-05-20',
+])
```
quantophred: Interesting to note the differences with [Indonesia's observance](https://github.com/quantopian/trading_calendars/pull/83/files?file-filters%5B%5D=.py#diff-b79ca87a50042e6e107c6c3567cff88fR370-R386), especially since the two countries are geographically pretty close. Calendars are hard!

PR #90, opened 2019-10-14
Add Pakistan Stock Exchange (XPSX)

quantophred:
Need to add more tests before we merge.


trading_calendars/exchange_calendar_xpsx.py
```diff
@@ -0,0 +1,394 @@
...
+    tz = timezone('Asia/Karachi')
+
+    open_times = (
+        (None, time(9, 32)),
+        # TODO: 09:17 on Fridays
+    )
+
+    close_times = (
+        (None, time(15, 30)),
+        # TODO: 16:30 on Fridays
```
dmichalowicz: Weird that they're open _longer_ on Fridays.

trading_calendars/exchange_calendar_xpsx.py
```diff
@@ -0,0 +1,394 @@
...
+    )
+
+    @property
+    def regular_holidays(self):
+        return HolidayCalendar([
```
dmichalowicz: Are you missing New Year's Day?

trading_calendars/exchange_calendar_xpsx.py
```diff
@@ -0,0 +1,394 @@
...
+                "Birthday of Quaid-e-Azam & Christmas",
+                month=12,
+                day=25,
+            ),
+        ])
+
+    juma_tul_wida = pd.to_datetime([
+        # Juma-Tul-Wida (last Friday of Ramadan before Eid-ul-Fitr)
+        '2004-11-11',  # NOTE: Observed on Thursday instead of Friday (??)
+        # Not observed in 2009 or 2010.
```
dmichalowicz: What about the other years between 2004 and 2013?

trading_calendars/exchange_calendar_xpsx.py
dmichalowicz: Never mind, I see you added it as adhoc below.

trading_calendars/exchange_calendar_xpsx.py
```diff
@@ -0,0 +1,394 @@
...
+        '2013-08-02',
+        '2014-07-25',
+        '2015-07-17',
+        '2016-07-01',
+        '2017-06-23',
+        '2018-06-08',
+        '2019-05-31',
+    ])
+
+    eid_ul_fitr = pd.to_datetime([
```
dmichalowicz: I see definitions of this holiday in other calendars (e.g. Indonesia) and also in `common_holidays.py`. Are they different enough that we have to redefine them for each country? It's fine if we do, but just thought I'd ask.

dmichalowicz: Also very minor, but we've called it `eid_al_fitr` elsewhere, not sure if there is an official spelling or anything, but maybe we want to be consistent everywhere.

trading_calendars/exchange_calendar_xpsx.py
```diff
@@ -0,0 +1,394 @@
...
+        '2018-06-15',  # Friday
+        '2018-06-18',  # Monday
+
+        '2019-06-04',  # Tuesday
+        '2019-06-05',
+        '2019-06-06',
+        '2019-06-07',
+    ])
+
+    eid_ul_azha = pd.to_datetime([
```
dmichalowicz: Same question here re spelling: `eid_ul_azha` vs. `eid_al_adha`.

tests/test_xpsx_calendar.py
```diff
@@ -0,0 +1,145 @@
...
+
+    answer_key_filename = 'xpsx'
+    calendar_class = XPSXExchangeCalendar
+
+    MAX_SESSION_HOURS = 5.967
+
+    HAVE_EARLY_CLOSES = False
+
+    # Pakistan observes Pakistan Standard Time (PKT, GMT+05:00) all year.
+    DAYLIGHT_SAVINGS_DATES = []
```
dmichalowicz: Instead of needing to overwrite this you can subclass this suite with `NoDSTExchangeCalendarTestBase`, which is defined in `test_trading_calendars.py`.

trading_calendars/exchange_calendar_xpsx.py
```diff
@@ -0,0 +1,394 @@
...
+    - Juma-Tul-Wida (last Friday of Ramadan)
+    - Eil-ul-Fitr (1st-3rd Shawwal)
+    - Eid-ul-Azha (10th-11th Zil-Hajj)
+    - Independence Day (Aug 14)
+    - Ashura (9th & 10th Muharram)
+    - Eid Milad-un-Nabi (12th Rabi-ul-Awal)
+    - Birthday of Quaid-e-Azam & Christmas (Dec 25)
+
+    Formerly-Observed Holidays:
+    - Iqbal Day (Nov 9; ended in 2012)
```
dmichalowicz: If you think it's worth it, may be good to add a quick test checking this date before/after 2012.

trading_calendars/exchange_calendar_xpsx.py
quantophred: The dates are pretty different everywhere I've seen. A lot of countries seem to base the start of these holidays on actual human observation of the crescent moon, so they're almost ad-hoc by definition, and bound to vary between exchanges.

I've been using the transliteration from each exchange's official list of market holidays. The spelling varies between exchanges (and sometimes even within a single exchange's website); but given that each exchange observes these holidays slightly differently, I think it's more important for each of our calendars to be consistent with the corresponding exchange than for our calendars to be consistent with each other. Not super strong on that, though -- thoughts?

(Once we drop Python 2 support, we can name it `عيد_الفطر` instead 😬)

tests/test_xpsx_calendar.py
quantophred: Ah, thanks. I'll make the change for Indonesia and South Africa too.

trading_calendars/exchange_calendar_xpsx.py
quantophred: It's actually split into two sessions on Friday: 9:17-12:00, and 2:32-4:30; so there's less total trading time. I should add a comment noting that, though. (We don't currently model intra-day closures, right?)

https://www.psx.com.pk/psx/exchange/general/trading-hours

trading_calendars/exchange_calendar_xpsx.py
```diff
@@ -0,0 +1,394 @@
...
+
+import pandas as pd
+from pandas.tseries.holiday import Holiday
+from pytz import timezone
+
+from .common_holidays import european_labour_day
+from .trading_calendar import HolidayCalendar, TradingCalendar
+
+
+class XPSXExchangeCalendar(TradingCalendar):
```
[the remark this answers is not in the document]
quantophred: Thanks -- I think I saw `psx.com.pk`, then confirmed that `XPSX` was present in the ISO file, but didn't notice that that's actually the code for `NASDAQ OMX PSX` (but not the  _operating_ code, which is `XNAS` 😓)

trading_calendars/exchange_calendar_xpsx.py
dmichalowicz: > We don't currently model intra-day closures, right?

Right.

trading_calendars/exchange_calendar_xpsx.py
quantophred: Good catch: looks like this is actually a *mostly* annual observance, but some of the other years had been mislabeled as Eid-ul-Fitr, or were unreported and found via analyzing the trading volume. Still not sure what happened in 2009 :\

tests/test_xkar_calendar.py
```diff
@@ -0,0 +1,98 @@
...
+                missing_sessions,
+                "\n{} weekday(s) were not listed as sessions by the"
+                " calendar, but were also not marked as holidays in the"
+                " test data:".format(len(missing_sessions))
+            )
+
+            self.fail("{} error(s) between {} and {}".format(
+                len(errors), start, end))
+
+    def test_iqbal_day(self):
```
dmichalowicz: 👍

PR #91, opened 2019-10-16
Argentina Calendar

peterhbromley:


trading_calendars/exchange_calendar_xbue.py
```diff
@@ -0,0 +1,373 @@
...
+
+SanMartinsDayTo2010 = Holiday(
+    "San Martin's Day",
+    month=8,
+    day=1,
+    end_date='2011',
+    offset=pd.DateOffset(weekday=MO(3)),
+)
+
+SanMartinsDayFrom2012 = Holiday(
```
dmichalowicz: Is this exactly the same as `SanMartinsDayTo2010`?

trading_calendars/exchange_calendar_xbue.py
```diff
@@ -0,0 +1,373 @@
...
+    day=6,
+    observance=not_2018,
+)
+
+DayOfNationalSovereigntyTo2016 = Holiday(
+    'Day of National Sovereignty (2010-2016)',
+    month=11,
+    day=1,
+    start_date='2010',
+    end_date='2015',
```
dmichalowicz: If it's `To2016` should this be '2017'?

trading_calendars/exchange_calendar_xbue.py
```diff
@@ -0,0 +1,373 @@
...
+    open_times = (
+        (None, time(11, 1)),
+    )
+
+    close_times = (
+        (None, time(17, 00)),
+    )
+
+    # TODO: Verify early close time
+    regular_early_close = time(13)
```
dmichalowicz: Is this still a TODO?

trading_calendars/exchange_calendar_xbue.py
dmichalowicz: Oh I see, it's skipping the year 2011. Might be clearer to use a `not_2011` observance like you've done with 2018? You could even generalize the observance function to be something like:

```python
def omit_year(year):
    """
    Return an observance function that excludes the given year from the holiday list.
    """
    def custom_observance(holidays):
        return holidays[holidays.year != year]
    return custom_observance
```

and then pass `observance=omit_year(2011)`, or 2018, etc.

trading_calendars/exchange_calendar_xbue.py
```diff
@@ -0,0 +1,373 @@
...
+            '2018-04-30',
+            '2018-12-24',
+            '2018-12-31',
+            '2019-07-08',
+            '2019-08-19',
+            '2019-10-14',
+        ]
+
+        irregular_observances = [
+            '2012-10-08',  # Cultural Diversity Day
```
dmichalowicz: Does this mean it was observed twice this year? Because the regular holiday doesn't omit 2012.

tests/test_xbue_calendar.py
```diff
@@ -0,0 +1,132 @@
+from unittest import TestCase
+import pandas as pd
+from pytz import UTC
+
+from .test_trading_calendar import NoDSTExchangeCalendarTestBase
+from trading_calendars.exchange_calendar_xbue import XBUEExchangeCalendar
+
+
+class XBUECalendarTestCase(NoDSTExchangeCalendarTestBase, TestCase):
```
dmichalowicz: From light googling I see that Argentina used to observe DST before 2009. What do you think of enabling the DST tests and overriding the dates to something pre-2009?

tests/test_xbue_calendar.py
```diff
@@ -0,0 +1,132 @@
...
+            pd.Timestamp('2019-05-24', tz=UTC),
+            pd.Timestamp('2019-05-27', tz=UTC),
+            # Martin Miguel de-Guemes Day on Sunday, Jun 17th.
+            pd.Timestamp('2018-06-15', tz=UTC),
+            pd.Timestamp('2018-06-18', tz=UTC),
+            # National Flag Day on Saturday, Jun 20th.
+            pd.Timestamp('2015-06-19', tz=UTC),
+            pd.Timestamp('2015-06-22', tz=UTC),
+            # Independence Day on Saturday, Jul 9th.
+            pd.Timestamp('2016-07-07', tz=UTC),
```
dmichalowicz: Did you mean `pd.Timestamp('2016-07-08', tz=UTC)` here?

tests/test_xbue_calendar.py
```diff
@@ -0,0 +1,132 @@
...
+        # non-Monday it is observed on the following Monday.  This means
+        # the holiday will be observed between Oct 10 and Oct 16.
+        expected_holidays = [
+            pd.Timestamp('2019-10-14', tz=UTC),  # Falls on Saturday
+            pd.Timestamp('2018-10-15', tz=UTC),  # Falls on Friday
+            pd.Timestamp('2017-10-16', tz=UTC),  # Falls on Thursday
+            pd.Timestamp('2016-10-10', tz=UTC),  # Falls on Wednesday
+            pd.Timestamp('2015-10-12', tz=UTC),  # Falls on Monday
+            pd.Timestamp('2014-10-13', tz=UTC),  # Falls on Sunday
+            pd.Timestamp('2013-10-14', tz=UTC),  # Falls on Saturday
```
dmichalowicz: Missing test for a Tuesday?

trading_calendars/exchange_calendar_xbue.py
peterhbromley: You're right, I think I forgot to rename this at some point!

trading_calendars/exchange_calendar_xbue.py
dmichalowicz: Oh right, I totally forgot about that.

trading_calendars/exchange_calendar_xbue.py
peterhbromley: Yeah it is, Jamie is unsure of the early close time so he said to just make it 1pm and make a note.  Does TODO work here or is there a different way I should be marking that?

trading_calendars/exchange_calendar_xbue.py
peterhbromley: The regular holiday actually does omit 2012, in `cultural_diversity_observance` the last line is filtering out 2012, which is why I also had to manually map `nearest_monday` over the holidays rather than just feed it directly to the observance parameter (which I thought was ugly/annoying but I wasn't sure how else to do it).

tests/test_xbue_calendar.py
peterhbromley: Looks like it was observed from 2007-2009.  I enabled the tests and added 2008 daylight savings dates

tests/test_xbue_calendar.py
peterhbromley: I had bumped it back a day because there was an adhoc on the 8th, but I decided to just change it to 2017 to avoid that issue.

tests/test_xbue_calendar.py
peterhbromley: Good idea, I added a Tuesday from 2010

trading_calendars/exchange_calendar_xbue.py
dmichalowicz: Ah missed that line 👍

PR #95, opened 2019-10-28
Holiday - Assembly Elections in Maharashtra

swapniljariwala:
There was a trading holiday on 21st October due to assembly elections in Maharashtra
https://www.bseindia.com/static/markets/marketinfo/listholi.aspx


jmccorriston:
Thanks for the PR, @swapniljariwala! I think all that's left to do is delete the Oct 21, 2019 row from the test file: https://github.com/quantopian/trading_calendars/blob/master/tests/resources/xbom.csv. If you make that change, I think the tests will pass. Alternatively, I can make a new PR with both changes. Unfortunately, I can't figure out how to append a new commit to this change 😕 .

swapniljariwala:
No worries, I will make the changes..

jmccorriston:
@swapniljariwala, we ended up adding in the missing holiday in [this PR](https://github.com/quantopian/trading_calendars/pull/103). As a result, I'm closing this PR. Thanks again for bringing up the issue!

PR #102, opened 2019-12-17
Test out gh actions for CI testing

gerrymanoim:


.github/workflows/main.yml
```diff
@@ -0,0 +1,46 @@
...
+    - uses: actions/checkout@v1
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      env:
+        PYTHONWARNINGS: ignore:DEPRECATION::pip._internal.cli.base_command
+      run: |
+        python -m pip install --upgrade pip
```
danisim: @gerrymanoim Doesn't look like installing dependencies takes that much time in the build but might want to consider adding this step prior https://github.com/actions/cache/blob/master/examples.md#python---pip?

gerrymanoim: Oh - thanks! I'll add that. 

Sadly the time is completely dominated by tests.

PR #108, opened 2020-01-03
added 2020 holidays to xkrx exchange

pkim81:


jmccorriston:
Thanks @pkim81! This matches the holidays published on the exchange website: http://global.krx.co.kr/contents/GLB/05/0501/0501110000/GLB0501110000.jsp. In order to merge the changes, we need the tests to pass, which means the tests need to be updated. Would you like me to update the tests, or do you want to try updating them yourself?

jmccorriston:
@pkim81 thanks for fixing the test, and thanks again for adding the 2020 holidays! Merging.

PR #116, opened 2020-01-18
Update setup.py

willianpaixao:
* README.md should be used as long description of the package so it can
be displayed in the PyPI website.
* Support to Python 3.4 was dropped, therefore removed from setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,12 +23,6 @@ import versioneer
 DISTNAME = "trading_calendars"
 DESCRIPTION = """trading_calendars is a Python library with \
 securities exchange calendars used by Quantopian's Zipline."""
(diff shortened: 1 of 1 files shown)


jmccorriston:
@willianpaixao Thanks for the PR! The README is a much better long description than what was hard coded in the setup.py file.

It looks like the Python 2.7 build is failing with `TypeError: 'encoding' is an invalid keyword argument for this function`. Since we still support a deprecated Python 2.7 environment on Quantopian, we'd like to keep supporting the Python 2.7 build for now. Is there any chance you could change the piece that reads the README so that the Py2 build passes as well?

willianpaixao:
@jmccorriston change applied.

jmccorriston:
Awesome, thanks @willianpaixao! LGTM

PR #121, opened 2020-02-05
Modify trading hours to suit ASX Single Closing Price Auction time + Good Friday early close 92-08

RichardDale:


ssanderson:
hey @RichardDale. Thanks for the PR. Can you provide a bit more context about what problem this is aiming to fix and why you think this is the correct fix?

RichardDale:
There's two parts to this commit... first is the trading hours - the Australian Securities Exchange actually has its single closing price auction at 4:11pm - so the 4pm close time needs to be amended.

Secondly, there used to be an early close prior to Easter.  This occurred from (at least) 1992 through 2008 inclusive.  

Both inclusions should be useful for people backtesting intraday data on Australian stocks.

ricpruss:
+1 on the close time problem.

jmccorriston:
@RichardDale - sorry for the slow reply and thanks for the info! Regarding the early closes, do you have a source that you could share either about the early closes, or about the change to remove the early closes after 2008?

Regarding the close time, I looked at the exchange site (https://www.asx.com.au/about/trading-hours.htm) and see this: "Between 4:00 pm and 4:10 pm, Sydney time, the market is placed in Pre-CSPA. Trading stops and brokers enter, change and cancel orders in preparation for the market closing". Can you help me understand what happens in this 10 minute window? Is one able to place new orders during this 10 minute period? My current (naive) interpretation is that this period is for brokers to settle their order books, but that might be totally wrong.

Issue #124, opened 2020-02-11
sessions_window with high negative count cycles back indices


pavitrakumar78:
Hi,

I've been experimenting with `sessions_window` and I found that when the `count` parameter is negative and greater than the total number of days in the trading calendar, the result cycles back and returns the dates between the end date and till the exact overflow value.

Example:
```
from trading_calendars import get_calendar
market_calendar = get_calendar('XNYS')
end_date = '2019-02-27'
days_rqd = -8500
dates = market_calendar.sessions_window(end_date, days_rqd)
print(dates)

>>DatetimeIndex(['2016-07-12 00:00:00+00:00', '2016-07-13 00:00:00+00:00',
               '2016-07-14 00:00:00+00:00', '2016-07-15 00:00:00+00:00',
               '2016-07-18 00:00:00+00:00', '2016-07-19 00:00:00+00:00',
               '2016-07-20 00:00:00+00:00', '2016-07-21 00:00:00+00:00',
               '2016-07-22 00:00:00+00:00', '2016-07-25 00:00:00+00:00',
               ...
               '2019-02-13 00:00:00+00:00', '2019-02-14 00:00:00+00:00',
               '2019-02-15 00:00:00+00:00', '2019-02-19 00:00:00+00:00',
               '2019-02-20 00:00:00+00:00', '2019-02-21 00:00:00+00:00',
               '2019-02-22 00:00:00+00:00', '2019-02-25 00:00:00+00:00',
               '2019-02-26 00:00:00+00:00', '2019-02-27 00:00:00+00:00'],
              dtype='datetime64[ns, UTC]', length=662, freq='C')

print(len(dates))
>> 662
```
Upon checking the code for this in trading_calendar.py, `start_idx` is 7345 for the calendar I've chosen in the above example and `end_idx` becomes 7345 + (-8500) = -1155 and thus the output will be `all_sessions[-1155: 7345 + 1]` which is not what I expected. The total length of `all_sessions` is 7839 - so, converting the negative indices to normal indices, we get  `all_sessions[7839-1155: 7345 + 1]` which is `all_sessions[6684: 7345 + 1]` which explains why I got 662 dates. 

[...]

PR #125, opened 2020-02-11
BUG: Do not loop-back when count is greater than historical dates

pavitrakumar78:
Made sure that `end_idx` in `sessions_window` method is always positive so that it does not loop-back when indexing on all_sessions when the count/lookback period is greater than the length of total historical dates. More details in #124.

### DIFF
--- a/trading_calendars/trading_calendar.py
+++ b/trading_calendars/trading_calendar.py
@@ -640,7 +640,7 @@ class TradingCalendar(with_metaclass(ABCMeta)):
         """
         start_idx = self.schedule.index.get_loc(session_label)
         end_idx = start_idx + count
-
+        end_idx = max(0, end_idx)
         return self.all_sessions[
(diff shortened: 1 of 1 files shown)


leonarduschen (approved):
Isn't this good to go?

Adding end_idx = max(0, end_idx) for handling cases where count << 0 is consitent with how the function handles cases where count >> 0 (i.e. just return the slice that satisfy the requirement and ignore the excess counts).

If we'd like to raise ValueError instead as @gerrymanoim suggested on #124 then IMO it'd have to be implemented both ways. More specifically, when end_idx < 0 and when end_idx > len(self.schedule). Let me know if this is the preferred behavior, I can create some new tests then make a separate PR for that :)

gerrymanoim (approved):

gerrymanoim:
Closes https://github.com/quantopian/trading_calendars/issues/124

PR #133, opened 2020-05-24
Fix pandas DeprecationWarning for empty series

asweeney86:
Initializing an empty pandas.Series without specifying a dtype will raise a
DeprecationWarning. Explicitly set the dtype to object for empty set
initialization.

> trading_calendar.py:983: DeprecationWarning: The default dtype for empty Series will be  'object' instead of 'float64' in a future version. Specify a dtype explicitly to silence this warning.
>    return pd.Series([])

### DIFF
--- a/trading_calendars/trading_calendar.py
+++ b/trading_calendars/trading_calendar.py
@@ -980,7 +980,7 @@ class TradingCalendar(with_metaclass(ABCMeta)):
         merged = regular + ad_hoc
         if not merged:
             # Concat barfs if the input has length 0.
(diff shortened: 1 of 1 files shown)


gerrymanoim:
Hey @asweeney86 - thank you for spotting this! 

I thinks, to be consistent with what this function returns in the non empty case, we should actually set the dtype as `datetime64[ns, UTC]`. 

```python
if not merged:
    return pd.Series([], dtype="datetime64[ns, UTC]")
````

PR #134, opened 2020-06-19
add break day because of the nCoV

lumyuto:


gerrymanoim:
@lumyuto - I don't see this holiday on https://www.tradinghours.com/exchanges/sse/market-holidays. Do you have a source?

lumyuto:
http://english.sse.com.cn/news/newsrelease/c/4993504.shtml

PR #141, opened 2020-08-14
Add temporary holiday to xkrx exchange

sjquant:
Korea has declared Aug 18, 2020 as temporary holiday, and the KRX market does not operate on that day.

See http://global.krx.co.kr/contents/GLB/05/0501/0501110000/GLB0501110000.jsp


gerrymanoim:
@sjquant - Thanks! Looks good to me.

Issue #145, opened 2020-09-02
Minor failures in all calendar tests


leonarduschen:
As the title suggests, there is always a minor failures in all calendar test `tests\test_XXXX_calendar.py` for all exchange code XXXX

For example, running `nosetests tests\test_xnys_calendar.py` yields

```
[success] 49.38% tests.test_xnys_calendar.XNYSCalendarTestCase.test_minute_to_session_label: 7.4921s
[success] 12.18% tests.test_xnys_calendar.XNYSCalendarTestCase.test_open_and_close_for_session: 1.8481s
[success] 11.56% tests.test_xnys_calendar.XNYSCalendarTestCase.test_next_prev_open_close: 1.7531s
[success] 11.36% tests.test_xnys_calendar.XNYSCalendarTestCase.test_next_prev_minute: 1.7233s
[success] 3.56% tests.test_xnys_calendar.XNYSCalendarTestCase.test_minute_index_to_session_labels_0: 0.5395s
[success] 3.05% tests.test_xnys_calendar.XNYSCalendarTestCase.test_next_prev_session: 0.4620s
[success] 2.26% tests.test_xnys_calendar.XNYSCalendarTestCase.test_sanity_check_session_lengths: 0.3434s
[success] 1.79% tests.test_xnys_calendar.XNYSCalendarTestCase.test_minute_index_to_session_labels_2: 0.2719s
[success] 1.78% tests.test_xnys_calendar.XNYSCalendarTestCase.test_is_open_on_minute: 0.2705s
[success] 1.73% tests.test_xnys_calendar.XNYSCalendarTestCase.test_minute_index_to_session_labels_1: 0.2623s
[success] 0.94% tests.test_xnys_calendar.XNYSCalendarTestCase.test_start_end: 0.1422s
[fail] 0.12% tests.test_xnys_calendar.XNYSCalendarTestCase.test_session_closes_in_range: 0.0187s
[success] 0.06% tests.test_xnys_calendar.XNYSCalendarTestCase.test_minutes_for_period: 0.0096s
[fail] 0.05% tests.test_xnys_calendar.XNYSCalendarTestCase.test_session_opens_in_range: 0.0069s
[success] 0.04% tests.test_xnys_calendar.XNYSCalendarTestCase.test_minutes_in_range: 0.0054s

Ran 27 tests in 15.773s

FAILED (failures=2)
```

gerrymanoim:
[...]

PR #146, opened 2020-09-02
Fix series equal assertion for issue 

leonarduschen:
This is a fix for issue #145 

It turns out the problem is that TradingCalendar.schedule datetimeindex has the frequency 'C' but the answer key does not because that info is lost when the data frame is saved as CSV.

Now `nosetests tests\test_xnys_calendar.py` yields

```
[success] 47.93% tests.test_xnys_calendar.XNYSCalendarTestCase.test_minute_to_session_label: 6.9389s
[success] 12.58% tests.test_xnys_calendar.XNYSCalendarTestCase.test_open_and_close_for_session: 1.8213s
[success] 11.51% tests.test_xnys_calendar.XNYSCalendarTestCase.test_next_prev_minute: 1.6660s
[success] 11.15% tests.test_xnys_calendar.XNYSCalendarTestCase.test_next_prev_open_close: 1.6139s
[success] 4.03% tests.test_xnys_calendar.XNYSCalendarTestCase.test_minute_index_to_session_labels_0: 0.5833s
[success] 3.11% tests.test_xnys_calendar.XNYSCalendarTestCase.test_next_prev_session: 0.4506s
[success] 2.35% tests.test_xnys_calendar.XNYSCalendarTestCase.test_sanity_check_session_lengths: 0.3397s
[success] 2.17% tests.test_xnys_calendar.XNYSCalendarTestCase.test_minute_index_to_session_labels_2: 0.3141s
[success] 1.99% tests.test_xnys_calendar.XNYSCalendarTestCase.test_is_open_on_minute: 0.2884s
[success] 1.83% tests.test_xnys_calendar.XNYSCalendarTestCase.test_minute_index_to_session_labels_1: 0.2648s
[success] 0.97% tests.test_xnys_calendar.XNYSCalendarTestCase.test_start_end: 0.1411s
[success] 0.06% tests.test_xnys_calendar.XNYSCalendarTestCase.test_minutes_for_period: 0.0089s
[success] 0.05% tests.test_xnys_calendar.XNYSCalendarTestCase.test_session_closes_in_range: 0.0078s
[success] 0.05% tests.test_xnys_calendar.XNYSCalendarTestCase.test_session_opens_in_range: 0.0077s
[success] 0.05% tests.test_xnys_calendar.XNYSCalendarTestCase.test_minutes_in_range: 0.0071s

Ran 27 tests in 15.080s
```

PR #148, opened 2020-09-02
BLD: Add PR labeler, bump pandas/numpy in testing

gerrymanoim:
Also closes https://github.com/quantopian/trading_calendars/issues/145

### DIFF
--- a/tests/test_trading_calendar.py
+++ b/tests/test_trading_calendar.py
@@ -784,7 +784,7 @@ class ExchangeCalendarTestBase(object):
             self.answers.index[0],
             self.answers.index[-1],
         )
-
+        found_opens.index.freq = None
         pd.util.testing.assert_series_equal(
             found_opens, self.answers['market_open']
         )
@@ -794,7 +794,7 @@ class ExchangeCalendarTestBase(object):
             self.answers.index[0],
             self.answers.index[-1],
         )
-
+        found_closes.index.freq = None
         pd.util.testing.assert_series_equal(
             found_closes, self.answers['market_close']
         )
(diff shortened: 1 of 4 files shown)


gerrymanoim:
Note - I may need to grab commits from https://github.com/quantopian/trading_calendars/pull/146. Will test locally.

leonarduschen (commented):
Yeah, tested locally and can confirm #146 pass for 1.1.1 as well (hence the false alarm in issue #145 )

tests/test_trading_calendar.py
```diff
@@ -784,7 +784,7 @@ def test_session_opens_in_range(self):
             self.answers.index[0],
             self.answers.index[-1],
         )
-
+        found_opens.index.freq = None
```
richafrank, line 787: Would we prefer instead to update the answer key to include the appropriate frequency? Not sure if that's easy.

richafrank (approved):
Looks good to me. Just had a suggestion.

tests/test_trading_calendar.py
gerrymanoim: I'm actually not convinced we want that. We're comparing our generated trading calendar (with a CustomBusinessDay frequency for that calendar) with an answer key of known good dates (but no frequency). I don't think we actually want to set the frequency of the answer key, since that's a property of the calendar logic we're testing. Newer pandas has `check_freq=False` as a parameter you pass, but that fails in pandas 18 (see https://github.com/quantopian/trading_calendars/pull/146/commits/96eb935a623af3779c2a7b4cf9b8ddf5ebc401c9). 

So we can either do a version check or set freq = None before doing the check.

PR #154, opened 2020-09-09
ENH: Add breaks to TradingCalendar

gerrymanoim:
Adds `break_start_times` and `break_end_times` to `TradingCalendar` to model intraday lunch breaks. These work as vectors of values, similarly to how `open_times` and `close_times` work so we can model break times changing over time. 

Threads the breaks through necessary methods. 

Adds breaks to XHKG. 

Items with `TODO` I'd especially like feedback on


gerrymanoim:
I _think_ all the pieces are wired appropriately now, just have to make the tests pass for holidays with breaks. The following are still broken:

- `test_minutes_in_range`
- `test_minutes_for_period`
- `test_next_prev_minute`

gerrymanoim:
FYI @rsheftel - breaks here should be functionally similar to `pandas_market_calendars` but work slightly differently internally.

rsheftel:
@gerrymanoim you may want to add the lunch break to the XSHG calendar as well. It is 11:30 AM - 1:00 PM Asia/Shanghai

gerrymanoim:
I was going to go through and add lunch breaks after the functional pieces of this PR were approved, happy to take care of it.

rsheftel:
@gerrymanoim makes sense. Once this lands on master I will look to add the CME specific calendars that require this feature.

trading_calendars/trading_calendar.py
```diff
@@ -1051,6 +1174,47 @@ def _overwrite_special_dates(midnight_utcs,
...
+
+    # -1 indicates that no corresponding entry was found.  If any -1s are
+    # present, then we have special dates that doesn't correspond to any
+    # trading day.
+    if -1 in indexer:
+        bad_dates = list(special_opens_or_closes[indexer == -1])
+        raise ValueError("Special dates %s are not trading days." % bad_dates)
+
+    # NOTE: This is a slightly dirty hack.  We're in-place overwriting the
+    # internal data of an Index, which is conceptually immutable.  Since we're
```
richafrank, line 1212: Is it important that we mutate this index, instead of creating a new one?

trading_calendars/trading_calendar.py
```diff
@@ -1051,6 +1174,47 @@ def _overwrite_special_dates(midnight_utcs,
     opens_or_closes.values[indexer] = special_opens_or_closes.values
 
 
+def _remove_breaks_for_special_dates(
+    midnight_utcs, break_start_or_end, special_opens_or_closes
+):
+    """
+    Overwrite breaks in break_start_or_end with corresponding dates in
+    special_opens_or_closes, using midnight_utcs for alignment.
```
richafrank, line 1182:
```suggestion
    Overwrite breaks in break_start_or_end on corresponding dates in
    special_opens_or_closes with NaT, using midnight_utcs for alignment.
```

trading_calendars/calendar_helpers.py
```diff
@@ -44,28 +64,43 @@ def is_open(opens, closes, minute_val):
...
-                      open_ + size * NANOSECONDS_PER_MINUTE,
-                      NANOSECONDS_PER_MINUTE)
-        )
-
-    out = np.concatenate(pieces).view('datetime64[ns]')
-    assert len(out) == num_minutes
+    NOTE: An extra minute is added to ending pbountaries (break_end and close)
+    so we include the last bar.
+    """
+    pieces = []  # todo preallocat?
```
richafrank, line 78: Regarding the todo, it's a performance question, so you could try that and compare.

trading_calendars/calendar_helpers.py
```diff
@@ -1,6 +1,9 @@
 import numpy as np
+import pandas as pd
 
 NANOSECONDS_PER_MINUTE = int(6e10)
+# TODO this is a negative number - can we just check that?
+NP_NAT = np.array([pd.NaT], dtype=np.int64)[0]
```
richafrank, line 6: A negative check seems reasonable to me, since the domain here is 24 hours. @llllllllll Thoughts on the call sites for this?

trading_calendars/calendar_helpers.py
```diff
@@ -25,14 +28,31 @@ def previous_divider_idx(dividers, minute_val):
...
     if open_idx != close_idx:
-        # if the indices are not same, that means the market is open
-        return True
+        # we are not guarenteed to have a sorted list of breaks, since break
+        # may not exist for a given date. Thus we need special logic.
+
+        # if the indices are not same, that means we are within a session
+        if (break_starts == NP_NAT).all() and (break_ends == NP_NAT).all():
+            # this calendar has no breaks
+            return True
```
richafrank, line 43: All these extra checks make me want us to build out breaks on every day, some off which may be null. Not sure the perf.

trading_calendars/calendar_helpers.py
```diff
@@ -25,14 +28,31 @@ def previous_divider_idx(dividers, minute_val):
...
+
+        break_start_on_open_dt = break_starts[open_idx - 1]
+        break_end_on_open_dt = break_ends[open_idx - 1]
+        if break_start_on_open_dt == NP_NAT:
+            # There is no break on the relevant day
+            return True
+        elif break_start_on_open_dt <= minute_val <= break_end_on_open_dt:
```
richafrank, line 50: Is the comparison correct here at the boundaries, given our labeling scheme?

trading_calendars/calendar_helpers.py
```diff
@@ -25,14 +28,31 @@ def previous_divider_idx(dividers, minute_val):
...
+            # we're in the middle of a break
+            return False
+        else:
+            return True
+
     else:
         try:
             # if they are the same, it might be the first minute of a
```
richafrank, line 58: What is this case? Does it apply to the re-open minute after the break?

trading_calendars/trading_calendar.py
```diff
@@ -224,8 +280,13 @@ def close_offset(self):
...
     @lazyval
     def _minutes_per_session(self):
-        diff = self.schedule.market_close - self.schedule.market_open
-        diff = diff.astype('timedelta64[m]')
+        close_to_open_diff = (
+            self.schedule.market_close - self.schedule.market_open
+        )
+        break_diff = (
+            self.schedule.break_end - self.schedule.break_start
+        ).fillna(pd.Timedelta(seconds=0))
```
richafrank, line 288: (Another case where I wonder if it would make sense to make break_end/break_start dense.)

richafrank (commented):
Assuming the boundary conditions and any other issues are all set, I'm a bit concerned about this as a behavior change for consumers (of `all_minutes`, `is_open_on_minute`, `minute_to_session_label`, etc.). Should we try this out with any tests downstream?

gerrymanoim:
> Should we try this out with any tests downstream?

What do you mean by that? There's a test in XHKG that tests if breaks are working correctly.

trading_calendars/trading_calendar.py
gerrymanoim: I'm not sure what you mean by make them dense? You could definitely have breaks that change over time (both over duration and whether they exist or not).

trading_calendars/exchange_calendar_xhkg.py
```diff
@@ -125,6 +126,12 @@ class XHKGExchangeCalendar(TradingCalendar):
         (None, time(10, 1)),
         (pd.Timestamp('2011-03-07'), time(9, 31)),
     )
+    break_start_times = (
+        (None, time(12, 0)),
+    )
+    break_end_times = (
+        (None, time(13, 0)),
```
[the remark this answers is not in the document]
gerrymanoim: Ah - you're right. 

IMO - given that `tranding_calendars` isn't strictly tied to backtesting anymore, the labeling scheme really isn't appropriate anymore. I naively would expect open/close/etc to reflect the actual opens and closes and for consuming applications to make adjustments as needed. 

I'll update this here for now, but would propose changing the scheme to be more natural in the next major version.

trading_calendars/trading_calendar.py
```diff
@@ -193,6 +229,26 @@ def open_times(self):
...
 
+    @property
+    def break_start_times(self):
+        """
+        Returns a list of tuples of (start_date, break_start_time).  If the
+        break start time is constant throughout the calendar, use None for the
+        start_date. If there is no break, use None for `break_start_time`.
+
+        """
+        return (None, None)
```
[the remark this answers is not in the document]
gerrymanoim: You're right - this should really be

```
return (
    (None, None)
)
```
matching  how one specifies open times (and what I did in `XKHG`.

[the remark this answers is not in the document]
gerrymanoim: I can switch the default to be `return None` though and modify accordingly.

trading_calendars/trading_calendar.py
```diff
@@ -114,6 +116,17 @@ def __init__(self, start=start_default, end=end_default):
             self.tz,
             self.open_offset,
```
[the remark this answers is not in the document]
gerrymanoim: The only place I see them is in the CME/IEPA/us_futures calendar. I can add them to breaks though.

trading_calendars/trading_calendar.py
```diff
@@ -129,15 +142,32 @@ def __init__(self, start=start_default, end=end_default):
...
         # information. This looks like it has been resolved in 0.17.1.
         # http://pandas.pydata.org/pandas-docs/stable/whatsnew.html#datetime-with-tz  # noqa
         self.schedule = DataFrame(
             index=_all_days,
-            columns=['market_open', 'market_close'],
+            columns=[
+                'market_open',
+                'break_start',
+                'break_end',
+                'market_close'
```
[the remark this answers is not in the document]
gerrymanoim: IIRC as of 3.something dicts are guarenteed to be ordered, they were ordered before, but that was an implementation detail. 


I think we will get the same result here actually not specifying `columns`. Unless that isn't true in older pandas?

trading_calendars/trading_calendar.py
```diff
@@ -1051,6 +1174,47 @@ def _overwrite_special_dates(midnight_utcs,
     opens_or_closes.values[indexer] = special_opens_or_closes.values
 
 
+def _remove_breaks_for_special_dates(
+    midnight_utcs, break_start_or_end, special_opens_or_closes
```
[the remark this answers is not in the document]
gerrymanoim: Will update that in `_overwrite_special_dates` as well.

trading_calendars/trading_calendar.py
gerrymanoim: This copies the logic we use in `_overwrite_special_dates`

trading_calendars/trading_calendar.py
```diff
@@ -861,13 +967,23 @@ def minute_to_session_label(self, dt, direction="next"):
...
-                           dt):
+            if not is_open(
+                self.market_opens_nanos,
+                self.market_break_starts_nanos,
+                self.market_break_ends_nanos,
+                self.market_closes_nanos,
+                dt
+            ):
                 # if the exchange is closed, use the previous session
                 return self.schedule.index[idx - 1]
```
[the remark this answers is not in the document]
gerrymanoim: Good point.

trading_calendars/trading_calendar.py
```diff
@@ -861,13 +967,23 @@ def minute_to_session_label(self, dt, direction="next"):
...
         elif direction == "none":
-            if not is_open(self.market_opens_nanos, self.market_closes_nanos,
-                           dt):
+            if not is_open(
+                self.market_opens_nanos,
+                self.market_break_starts_nanos,
+                self.market_break_ends_nanos,
+                self.market_closes_nanos,
+                dt
```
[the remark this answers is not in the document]
gerrymanoim: I think they should? We can make this a major version bump, but if I'm passing in `minute_to_session_label` and specifically using `direction="none"` I would expect a `ValueError` if the exchange is closed.

trading_calendars/exchange_calendar_xhkg.py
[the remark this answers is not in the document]
rsheftel: For my two-cents I would suggest open, close, break-start and break-end be the times of those events, and not rounded up to the next minute. [this is probably a separate discussion bigger than just the break functionality] 

My rationale is that (1) that is the market norm and what people would expect when looking at the calendars, people say that NYSE opens at 9:30am, not 9:31am (2) adding a minute to the open won't work for frequencies less than a minute and (3) some system will want to do something on the open, so having that in the date_range() function is required.

If someone wants to get the end of the first bar after the open, the pandas date_range() provides the "closed" parameter for this purpose.

trading_calendars/trading_calendar.py
```diff
@@ -755,14 +821,35 @@ def open_and_close_for_session(self, session_label):
         (Timestamp, Timestamp)
             The open and close for the given session.
         """
-        sched = self.schedule
+        # `market_open` and `market_close` should be timezone aware, but pandas
```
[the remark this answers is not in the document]
gerrymanoim: This comment is no longer relevant, I believe the underlying issue was fixed in pandas 17.1 (see message at https://github.com/quantopian/trading_calendars/blob/d915eaa86557e8aee1b61f100df9d7e6ea080e8d/trading_calendars/trading_calendar.py#L155). 

The reason this dataframe has no tz info on it is we specifically pass a `dtype` argument in our `self.schedule` construction. 

```python
self.schedule = DataFrame(
            index=_all_days,
            columns=[
                'market_open',
                'break_start',
                'break_end',
                'market_close'
            ],
            data={
                'market_open': self._opens,
                'break_start': self._break_starts,
                'break_end': self._break_ends,
                'market_close': self._closes,
            },
            dtype='datetime64[ns]',
        )
```

I'm going to remove the comment here and put up a seperate PR dropping that `dtype` so we don't have to `tz_localize(UTC)` everywhere.

trading_calendars/calendar_helpers.py
gerrymanoim: This is now handled in the check above:

```
if break_start_on_open_dt <= minute_val < break_end_on_open_dt:
            # we're in the middle of a break
            return False
        else:
            return True
```

If `minute_val==break_end_on_open_dt` we'll return `True`.

trading_calendars/trading_calendar.py
```diff
@@ -340,7 +398,7 @@ def is_session(self, dt):
         """
         return dt in self.schedule.index
 
-    def is_open_on_minute(self, dt):
+    def is_open_on_minute(self, dt, ignore_breaks=False):
```
richafrank, line 401: Shall we add some test coverage for `ignore_breaks=True` and add it to the docstring here?

trading_calendars/trading_calendar.py
richafrank: Ah, nevermind. This makes sense now.

trading_calendars/exchange_calendar_xhkg.py
```diff
@@ -92,6 +92,7 @@ class XHKGExchangeCalendar(TradingCalendar):
     Exchange calendar for the Hong Kong Stock Exchange (XHKG).
 
     Open Time: 9:31 AM, Asia/Hong_Kong
+    Lunch Break: 12:00 PM - 1:00 PM Asia/Hong_Kong
```
richafrank, line 95: At least until we change our start time labels as @rsheftel was suggesting, shall we make this consistent?
```suggestion
    Lunch Break: 12:01 PM - 1:00 PM Asia/Hong_Kong
```

trading_calendars/calendar_helpers.py
```diff
@@ -25,14 +28,25 @@ def previous_divider_idx(dividers, minute_val):
...
+def is_open(opens, break_starts, break_ends, closes, minute_val):
 
     open_idx = np.searchsorted(opens, minute_val)
     close_idx = np.searchsorted(closes, minute_val)
 
     if open_idx != close_idx:
-        # if the indices are not same, that means the market is open
-        return True
+        # we are not guaranteed to have a sorted list of breaks, since break
+        # may not exist for a given date. Thus we need special logic.
```
richafrank, line 38: Is this still true? I think we have values for all breaks now (some are NaT).

trading_calendars/calendar_helpers.py
```diff
@@ -44,28 +58,63 @@ def is_open(opens, closes, minute_val):
...
-    deltas = closes_in_ns - opens_in_ns
-
-    # + 1 because we want 390 mins per standard day, not 389
-    daily_sizes = (deltas // NANOSECONDS_PER_MINUTE) + 1
-    num_minutes = daily_sizes.sum()
+    open_idx = np.searchsorted(opens, minute_val)
+    close_idx = np.searchsorted(closes, minute_val)
+    if open_idx != close_idx:
+        # if the indices are not same, that means we are within a session
+        return True
```
richafrank, line 69: This looks correct to me, but what do you think about merging this function back into `is_open` and giving it an `ignore_breaks` parameter, so that we're not duplicating the non-trivial logic of `is_open`?

This different block would turn into:
```python
if ignore_breaks:
    return True
else:
    # break-specific logic above
```

trading_calendars/trading_calendar.py
```diff
@@ -812,20 +920,16 @@ def all_minutes(self):
...
+            compute_all_minutes(
+                self.market_opens_nanos,
+                self.market_break_starts_nanos,
+                self.market_break_ends_nanos,
+                self.market_closes_nanos,
+            ),
             tz=UTC,
         )
 
-    @preprocess(dt=coerce(pd.Timestamp, attrgetter('value')))
```
richafrank, line 828: If we remove the preprocessor, then do we need to check whether `dt` is a `Timestamp`, since it could be an int?
```
        Parameters
        ----------
        dt : pd.Timestamp or nanosecond offset
            The dt for which to get the containing session.
```

If we do remove this lone `preprocess` call site, then I think we can actually kill the entire preprocess module.

tests/test_trading_calendar.py
```diff
@@ -571,10 +571,28 @@ def test_minutes_for_period(self):
         _open, _close = self.calendar.open_and_close_for_session(
             full_session_label
         )
+        _break_start, _break_end = (
+            self.calendar.break_start_and_end_for_session(
+                full_session_label
+            )
+        )
+        if _break_start is not pd.NaT:
```
richafrank, line 579:
```suggestion
        if not pd.isnull(_break_start):
```

tests/test_trading_calendar.py
```diff
@@ -679,23 +697,48 @@ def test_minutes_in_range(self):
...
+
+        intervals = [
+            (first_open, first_break_start, first_break_end, first_close),
+            (middle_open, middle_break_start, middle_break_end, middle_close),
+            (last_open, last_break_start, last_break_end, last_close),
+        ]
+        all_minutes = []
+
+        for _open, _break_start, _break_end, _close in intervals:
+            if _break_start is pd.NaT:
```
richafrank, line 718:
```suggestion
            if pd.isnull(_break_start):
```

richafrank (commented):
Nice improvements! Added some suggestions. The preprocess change looked like the only breaking change.

trading_calendars/trading_calendar.py
```diff
@@ -129,15 +141,32 @@ def __init__(self, start=start_default, end=end_default):
         # Overwrite the special opens and closes on top of the standard ones.
         _overwrite_special_dates(_all_days, self._opens, _special_opens)
         _overwrite_special_dates(_all_days, self._closes, _special_closes)
+        _remove_breaks_for_special_dates(
+            _all_days,
+            self._break_starts,
+            _special_closes,
```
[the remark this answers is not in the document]
gerrymanoim: I'm not sure - I've seen a bunch of examples where if there's a special close (half day) there's no mid day break. I don't know what the logic with late opens might be (those seem to be far rarer/due to tech glitches).

trading_calendars/trading_calendar.py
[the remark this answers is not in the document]
gerrymanoim: Only reason I don't want to use an `OrderedDict` is because we still(?) want to support py27 and that was introduced in py 3.2. I'd rather not add a compatibility check here.

trading_calendars/calendar_helpers.py
```diff
@@ -44,28 +64,43 @@ def is_open(opens, closes, minute_val):
...
-    out = np.concatenate(pieces).view('datetime64[ns]')
-    assert len(out) == num_minutes
+    NOTE: An extra minute is added to ending pbountaries (break_end and close)
+    so we include the last bar.
+    """
+    pieces = []  # todo preallocat?
+    for open_time, break_start_time, break_end_time, close_time in zip(
+        opens_in_ns, break_starts_in_ns, break_ends_in_ns, closes_in_ns
+    ):
+        if break_start_time != NP_NAT and break_end_time != NP_NAT:
```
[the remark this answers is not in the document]
gerrymanoim: I don't think this will have performance implications, but I'll add a check at calendar construction that will raise an error if you have unmatched break starts/ends. It will be clearer.

trading_calendars/trading_calendar.py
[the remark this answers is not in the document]
richafrank: I'm very happy leaving this as is, but just so you're aware, `OrderedDict` is `New in version 2.7.`

trading_calendars/trading_calendar.py
```diff
@@ -850,24 +973,26 @@ def minute_to_session_label(self, dt, direction="next"):
...
-            self._minute_to_session_label_cache = (dt, current_or_next_session)
+            self._minute_to_session_label_cache = (
+                dt.value, current_or_next_session
+            )
             return current_or_next_session
         elif direction == "previous":
-            if not is_open(self.market_opens_nanos, self.market_closes_nanos,
-                           dt):
-                # if the exchange is closed, use the previous session
+            if not self.is_open_on_minute(dt, ignore_breaks=True):
```
richafrank, line 992: I notice that we convert int->Timestamp above, then we use only `.value` everywhere here and pass the timestamp to `is_open_on_minute`, which also uses only `.value`. Would it make sense to have both these methods do a Timestamp->int conversion at their starts, and then work with ints?

richafrank (commented):
Nice! Just had one suggestion.

trading_calendars/trading_calendar.py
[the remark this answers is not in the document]
gerrymanoim: My bad - I misread the docs here.

Issue #152, opened 2020-09-04
Data Differences Between `trading_calendars` and `pandas_market_calendars`


gerrymanoim:
re https://github.com/rsheftel/pandas_market_calendars/issues/41

# Summary status

- [x] XHKG
   - https://github.com/quantopian/trading_calendars/issues/152#issuecomment-689799967
- [x] CFE
   - https://github.com/quantopian/trading_calendars/issues/152#issuecomment-689812670
   - Note: We don't know what to do about `1994-04-27` in `pmc` but not in `tc`
- [ ] XSWX
- [X] LSE
   - Issue: https://github.com/quantopian/trading_calendars/issues/163
   - PR: https://github.com/quantopian/trading_calendars/pull/166
- [ ] XSHG
- [ ] XASX
- [X] TSX
   - https://github.com/quantopian/trading_calendars/issues/152#issuecomment-689136700) 
- [ ] BMF

## Extras not covered by diff below:

- [x] Pre 1990 NYSE holidays
   - https://github.com/quantopian/trading_calendars/issues/162

# Details

**code generating the table below**

```python
from collections import namedtuple 

import pandas_market_calendars as pmc
import trading_calendars as tc

# todo port this back to trading calendars
def get_tc_calendar_names():
  return list(tc.calendar_utils._default_calendar_factories.keys())+list(tc.calendar_utils._default_calendar_aliases.keys())

tc.get_calendar_names = get_tc_calendar_names
# end todo

pmc_calendars = set(pmc.get_calendar_names())
tc_calendars = set(tc.get_calendar_names())
tc_to_pmc_mapping = {
    'XASX': 'ASX',
    'XHKG': 'HKEX',
    'XTKS': 'JPX',
    'XOSL': 'OSE',
    'XSWX': 'SIX',
    'XSHG': 'SSE',
}

in_both = (tc_calendars & pmc_calendars) | set(tc_to_pmc_mapping.keys())

CompareResult = namedtuple("Result", ["only_tc", "only_pmc"])

def compare_calendars(cal: str):
  tc_cal = tc.get_calendar(cal).all_sessions
  pmc_cal_name = tc_to_pmc_mapping.get(cal, cal)
  pmc_cal = pmc.get_calendar(pmc_cal_name).valid_days(min(tc_cal), max(tc_cal))
  if not (tc_cal ^ pmc_cal).empty:
    return CompareResult

out = {}
for cal in in_both:
  diff = compare_calendars(cal)
  if diff:
[...]

PR #165, opened 2020-09-16
ENH: Update XHKG calendar

gerrymanoim:
WIP PR for updating XHKG. See details here: https://github.com/quantopian/trading_calendars/issues/152#issuecomment-689799967.

Pausing for now as:

- I need to do a bit more refactoring to get Buddha's birthday/other lunar holidays ported over correctly from `pandas_market_calendars`
- I need clarification from @jmccorriston on `Liberation Day` (I can't see that in either calendar?)
- I need some other clarification about whether we're pulling the rest of the missing holidays from PMC (which I believe to be the case, but I wanted to double check).


jmccorriston:
Liberation Day is another name for Allied Victory Day https://en.wikipedia.org/wiki/Victory_over_Japan_Day, which is currently implemented here in pmc: https://github.com/rsheftel/pandas_market_calendars/blob/master/pandas_market_calendars/exchange_calendar_hkex.py#L240.

In general, I think we should port over the rest of the holidays from pmc.

gerrymanoim:
@jmccorriston - updated, I think I grabbed them all, will run another check when I have a sec.

gerrymanoim:
Confirmed no more diffs between the two calendars.

jmccorriston:
Sorry for the delay, this lgtm then!

jmccorriston (approved):

Issue #162, opened 2020-09-15
NYSE Calendar differences prior to 1990


rsheftel:
I see that the earliest starting date for the NYSE (XNYS) calendar is 1990-01-01 and from that date onwards the trading_calendars and pandas_market_calendars implementation are the same. But prior to that date there are many changes in the holiday calendar that pcal has that trading_calendars does not. If you are OK, I would suggest bringing over the pre-1990 holidays. 

These holidays were added by @pldrouin in https://github.com/rsheftel/pandas_market_calendars/pull/30. He was nice enough to upload a csv of all the holidays in the /tests/data/nyse_all_full_day_holidays_since_1928.csv and the results can be verified here: http://s3.amazonaws.com/armstrongeconomics-wp/2013/07/NYSE-Closings.pdf

Example:

trading_calendars:
```
from .us_holidays import (
    USNewYearsDay,
    USMartinLutherKingJrAfter1998,
    USMemorialDay,
    USIndependenceDay,
    Christmas,
    MonTuesThursBeforeIndependenceDay,
    FridayAfterIndependenceDayExcept2013and2019,
    USBlackFridayBefore1993,
    USBlackFridayInOrAfter1993,
    September11Closings,
    HurricaneSandyClosings,
    HurricaneGloriaClosing,
    NewYorkBlackout,
    PresidentialElectionDays,
    USNationalDaysofMourning,
    ChristmasEveBefore1993,
    ChristmasEveInOrAfter1993,
)
```

pandas_market_calendars
 ```
def regular_holidays(self):
        return AbstractHolidayCalendar(rules=[
            USNewYearsDay,
            USMartinLutherKingJrAfter1998,
            USLincolnsBirthDayBefore1954,
            USWashingtonsBirthDayBefore1964,
            USWashingtonsBirthDay1964to1970,
            USPresidentsDay,
            GoodFriday,
            USMemorialDayBefore1964,
            USMemorialDay1964to1969,
            USMemorialDay,
            USIndependenceDayBefore1954,
            USIndependenceDay,
            USLaborDay,
            USThanksgivingDayBefore1939,
            USThanksgivingDay1939to1941,
            USThanksgivingDay,
            USElectionDay1848to1967,
            USElectionDay1968to1980,
            USVeteransDay1934to1953,
            USColumbusDayBefore1954,
            ChristmasBefore1954,
            Christmas,
        ])
```

gerrymanoim:
[...]

PR #169, opened 2020-09-22
ENH: NYSE Calendar prior to 1990

gerrymanoim:
Closes https://github.com/quantopian/trading_calendars/issues/162

I didn't update the tests to go back to 1928. I wanted to make sure our tests still passed and I'm not sure testing that far back is worth it, but definitely what to hear what you all think. 

If you're looking at the diff, it makes sense to look at the individual commits. I ran a formatter on the `.py` files because the code formatting was all over the place.


jmccorriston:
This change LGTM. I don't think we need to add tests pre-1990 but maybe we can add a disclaimer (or a runtime warning?) saying that holidays prior to 1990 have not been corroborated with volume data and are not included in the unit tests?

jmccorriston (approved):

rsheftel:
I have run this through the test suite in pandas_market_calendars that covers the pre-1990 and all tests pass. At some point I can add those tests here.

PR #173, opened 2020-10-08
TST: Add py3.9 to test matrix

gerrymanoim:
Drop 3.7.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions" # See documentation for possible values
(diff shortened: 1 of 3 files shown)


gerrymanoim:
Ack - looks like osx 3.9 numpy wheels still aren't quite available.

gerrymanoim:
@richafrank So this passes CI now, but we do have a mild performance regression! Cool to see that CI in action.

richafrank:
> So this passes CI now, but we do have a mild performance regression! Cool to see that CI in action.

So nice that we can see that! Anything obvious we should change in our numpy usage, or shall we just accept the regression?

richafrank:
(The diff itself looks reasonable to me.)

gerrymanoim:
There was nothing obvious in the release notes: https://github.com/numpy/numpy/releases/tag/v1.19.3. 

I was going to do some mild profiling locally, but merge this otherwise if there was nothing obvious.

richafrank:
Sounds good to me.

gerrymanoim:
Profiling calendar init, the difference in time is in `_all_days = date_range(start, end, freq=self.day, tz=UTC)` during the building of the calendar. 

Going to merge this as that's all outside our code and I still is a reasonable thing to do.

Issue #160, opened 2020-09-14
TST: Move off of nosetests


gerrymanoim:
And on to nose2 or pytest

PR #179, opened 2020-10-13
TST: Move off nose and on to pytest

gerrymanoim:
Closes https://github.com/quantopian/trading_calendars/issues/160

I didn't change any of the `self.assert*` code or move this from a class interface to a more `pytset.fixtures` interface. I'd like to do that after py2 gets dropped.

### DIFF
--- a/etc/requirements.txt
+++ b/etc/requirements.txt
@@ -2,9 +2,7 @@
 configparser==3.5.0
 enum34==1.1.6
 flake8==3.7.9
-nose==1.3.7
-nose-ignore-docstring==0.2
-nose-timer==0.7.5
+pytest==6.0.1
 parameterized==0.6.1
 python-dateutil==2.7.3
 pytz==2019.3
--- a/etc/requirements.txt
+++ b/etc/requirements.txt
@@ -2,7 +2,8 @@
 configparser==3.5.0
 enum34==1.1.6
 flake8==3.7.9
-pytest==6.0.1
+pytest==4.6.11;python_version<"3.5"
+pytest==6.0.1;python_version>"2.7"
 parameterized==0.6.1
 python-dateutil==2.7.3
 pytz==2019.3
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,13 +1,3 @@
-[nosetests]
-verbosity=2
-detailed-errors=1
-with-ignore-docstrings=1
-with-timer=1
-timer-top-n=15
-cover-package=trading_calendars
-logging-level=INFO
-traverse-namespace=1
-
 [metadata]
 description-file = README.md
 license_file = LICENSE
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,7 +2,7 @@
 description-file = README.md
 license_file = LICENSE
 
-[tools:pytest]
+[tool:pytest]
 addopts = -v --durations=15
 
 # See the docstring in versioneer.py for instructions. Note that you must
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,6 +2,9 @@
 description-file = README.md
 license_file = LICENSE
 
+[tools:pytest]
+addopts = -v --durations=15
+
 # See the docstring in versioneer.py for instructions. Note that you must
 # re-run 'versioneer.py setup' after changing this section, and commit the
 # resulting files.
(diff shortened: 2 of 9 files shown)


gerrymanoim:
Confirmed that the overall test count is the same in both cases: 1254.

etc/requirements.txt
```diff
@@ -2,9 +2,8 @@
 configparser==3.5.0
 enum34==1.1.6
 flake8==3.7.9
-nose==1.3.7
-nose-ignore-docstring==0.2
-nose-timer==0.7.5
+pytest==4.6.11;python_version<"3.5"
+pytest==6.0.1;python_version>"2.7"
```
richafrank, line 6: Am I right that this file is meant to be a lockfile? If so, then we should pin the transitive dependencies too.

setup.cfg
```diff
@@ -1,13 +1,3 @@
-[nosetests]
-verbosity=2
```
richafrank, line 2: What do you think about running pytest with some verbosity? I usually find it useful.

setup.cfg
```diff
@@ -1,13 +1,3 @@
...
-detailed-errors=1
-with-ignore-docstrings=1
-with-timer=1
-timer-top-n=15
-cover-package=trading_calendars
```
richafrank, line 7: Do we still have coverage running at all?

richafrank (commented):
Nice!

setup.cfg
gerrymanoim: I normally set that when I'm trying to debug something and am testing locally. Happy to change the default if you think that's better.

etc/requirements.txt
gerrymanoim: Not really sure. This is a relic from the appveyor/travis days that we seem to have never touched. I can convert this to a pip-tools flow, but I'd like to do that as a separate PR if that's okay.

tests/test_calendar_dispatcher.py
```diff
@@ -14,7 +14,7 @@
 class CalendarAliasTestCase(TestCase):
 
     @classmethod
-    def setupClass(cls):
+    def setup_class(cls):
```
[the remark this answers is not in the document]
gerrymanoim: I think it would have - but this was a simple change to make. Happy to change it back.

setup.cfg
```diff
@@ -1,13 +1,3 @@
-[nosetests]
-verbosity=2
-detailed-errors=1
-with-ignore-docstrings=1
-with-timer=1
```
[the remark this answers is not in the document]
gerrymanoim: I'll update the duration. I don't think we've ever done anything with this though and I generally do not look through/compare the timings of individual tests. 

If we care about performance regressions, we should add something like https://github.com/rhysd/github-action-benchmark (using https://pypi.org/project/pytest-benchmark/). Happy to work on getting benchmark CI running in a separate PR.

setup.cfg
gerrymanoim: I don't see where we were running or reporting on this information. Happy to run something under https://pypi.org/project/pytest-cov/ or https://coverage.readthedocs.io/en/latest/. 

If we do run coverage somewhere I'd like to report it somewhere actionable, not just as a log into the test suite.

richafrank: Maybe it's a relic from zipline, where we are or were using coveralls. Yep, feel free to add later, if we think it's useful.

setup.cfg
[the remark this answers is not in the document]
richafrank: Ooo that would be nice.

setup.cfg
richafrank: Yea, that'd be my preference.

etc/requirements.txt
richafrank: Sounds good.

tests/test_calendar_dispatcher.py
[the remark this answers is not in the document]
gerrymanoim: Yep - just consistency. Sesemed easy enough to fix quickly.

setup.cfg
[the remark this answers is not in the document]
gerrymanoim: Follow up: https://github.com/quantopian/trading_calendars/issues/180

etc/requirements.txt
gerrymanoim: https://github.com/quantopian/trading_calendars/issues/181

setup.cfg
gerrymanoim: Follow up: https://github.com/quantopian/trading_calendars/issues/182

Issue #180, opened 2020-10-13
Add continuous benchmarking


gerrymanoim:
We should add something like https://github.com/rhysd/github-action-benchmark (using https://pypi.org/project/pytest-benchmark/).

PR #184, opened 2020-10-13
BENCH: Add CI benchmarking and alerting on regressions

gerrymanoim:
Closes https://github.com/quantopian/trading_calendars/issues/180

cc @llllllllll if you want to use this for anything. 

Right now I'm testing the constructuction of all of our calendars (which is our slowest test), but I'm open to other suggestions of what we should alert on.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -15,7 +15,7 @@ jobs:
         python-version: 3.8
     - name: Install dependencies
       run: |
-        pip install -e .[test] -c etc/requirements_locked.txt
+        pip install -e .[dev] -c etc/requirements_locked.txt
     - name: Run benchmark
       run: |
         pytest etc/bench.py --benchmark-json output.json
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -35,4 +35,4 @@ jobs:
         github-token: ${{ secrets.GITHUB_TOKEN }}
         # Enable alert commit comment
         comment-on-alert: true
-        alert-threshold: "125%"
+        alert-threshold: "110%"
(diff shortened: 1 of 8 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,38 @@
+name: Benchmark
+
+on:
+  pull_request:
+    branches:
+      - master
```
richafrank, line 6: Not specific to this PR, but should we run CI/benchmark on pushes to master as well?

etc/bench.py
```diff
@@ -0,0 +1,27 @@
...
+        _default_calendar_aliases, _default_calendar_factories
+    )
+
+    for name in calendar_names:
+        assert dispatcher.get_calendar(name) is not None
+        dispatcher.deregister_calendar(name)
+
+
+def test_calendar_construction(benchmark):
+    benchmark(construct_all_calendars)
```
richafrank, line 27: Should we benchmark just the loop above that does the construction (as opposed to the dispatcher setup, etc, as well)?

richafrank (commented):
This is exciting! From looking at test timings and loops (PR-pending), I'd recommend we benchmark `minute_to_session_label` as well.

etc/bench.py
gerrymanoim: I don't think the overhead of dict construction is significant relative to constructing all the calendars. It would be more accurate, but I don't think would give you a different answer.

.github/workflows/benchmark.yml
gerrymanoim: I can add CI to master mergers - I was hesitant at first because the CI here actually takes a long time. 

Not sure it is worth benchmarking the master merge as well.

etc/bench.py
richafrank: Sounds fine then.

PR #185, opened 2020-10-14
Added missing direct dependencies to install_requires

richafrank:
Found in https://github.com/conda-forge/trading-calendars-feedstock/pull/1

### DIFF
--- a/etc/requirements.in
+++ b/etc/requirements.in
@@ -1,4 +1,6 @@
 numpy
 pandas
+python-dateutil
 pytz
+six
 toolz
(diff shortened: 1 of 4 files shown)


gerrymanoim:
I actually don't think we should have `-e` as part of the requirements.

gerrymanoim:
I'd lean to either (c) or (a) - apologies, should have mentioned that in my original comment.

richafrank:
Ok, let's go with (a) the simplest choice. If it ever becomes an actual burden to update both locations, we can rethink it.

gerrymanoim (approved):
LGTM - thanks!

PR #186, opened 2020-10-15
TST: Improve our CI tests speed

richafrank:
- Uses faster pandas methods for iteration
- Parallelizes tests across CPUs
- Sped up calendar construction for special opens and closes
- Removed a deprecation warning

An example (macos-latest, 3.8):
"Test with pytest" (before): 22m36s
"Test with pytest" (after): 3m54s

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,4 +56,4 @@ jobs:
         flake8
     - name: Test with pytest
       run: |
-        pytest tests
+        pytest tests -n auto --dist loadscope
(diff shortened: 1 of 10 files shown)


richafrank:
@gerrymanoim Looking forward to rebasing this after your benchmark merge.

gerrymanoim:
> The cache is scoped to the key and branch. The default branch cache is available to other branches.

TIL - guess we should run the benchmark on master as well so this gets cached correctly.

Issue #129, opened 2020-03-12
Printing Trading Calendar


ardita96:
Hi there, 

I get the following error when trying to print the calendar, how can I view it?
<trading_calendars.exchange_calendar_xnys.XNYSExchangeCalendar object at 0x00000000105AF1C0>
Thanks

oyvindwe:
Looks like the opening hours for XETR is those that are implemented. Here is the official source: https://www.xetra.com/xetra-en/trading/trading-calendar-and-trading-hours

XFRA opens 08:00 and closes 20:00 CET for most products.

If XFRA is changed, I think XETR should be added with 09:00-17:30 CET and same calendar.

jenskeiner:
Besides the issue with the regular business hours, the closing time for the special close on 30th December seems to be incorrect as well, regardless of whether you consider XFRA or XETR.

Consider e.g. the [official announcement for 2019](https://www.deutsche-boerse.com/resource/blob/1665538/2e52afc0758f13724115da34f66e1df8/data/Bekanntmachung-zur-Jahresschlussb-rse-2019.pdf) according to which XETR closed the regular session at 14:00 CET (with a post-trading phase until 15:00 CET) and XFRA closed at 15:00 CET. The time returned by the respective calendar in the package at version 1.11.8 is 12:30 CET.

So it would appear that the special close time for XFRA should be 15:00 CET and 14:00 CET for XETR since the post-trading phase is typically not considered.

Also of note is that this special close is decided upon every year again, so to be 100% accurate, one would have watch out for the official announcement in December and update the code in case the expected special times don't apply.

I would suggest to add XETR as this would allow to fix the times for good. Otherwise, you'll always be breaking one or the other.

PR #197, opened 2020-12-07
Add XETR as a separate market.

jenskeiner:
See discussion in #129. XFRA and XETR, while both part of Deutsche Börse AG, are actually two different markets represented by different operating MICs. They largely share the same properties, but exchange hours are different.

This PR is to add a separate calendar for XETR.

Compared to XFRA, I've adjusted the early close time to 14:00 CET (was 12:30 for XFRA, not entirely sure this is actually correct). Everything else already seemed to be consistent.

I've added a note in the code that the early close day towards the end of the year is decided upon separately each year in December. So, having a rule for this is basically just a working assumption of what will happen in the future as well.

With this merged in, XFRA could be adjusted independently if, according to #129, exchange times are indeed incorrect there.


gerrymanoim:
Thanks - LGTM. 

@richafrank

gerrymanoim (approved):

PR #199, opened 2020-12-09
Fix XTKS 2021 holidays according to JPX calendar

sercant:
Updated according to JPX 2021 Market Holidays.

Ref: https://www.jpx.co.jp/english/corporate/about-jpx/calendar/


gerrymanoim:
This is great, thanks! LGTM.

gerrymanoim (approved):

PR #210, opened 2021-01-07
ENH: Add rules for XTAE holidays.

jenskeiner:
This PR adds proper rules for regular XTAE holidays as opposed to just having everything defined as ad-hoc holidays.

I've done my best to cross-check this against the official information here: https://info.tase.co.il/eng/about_tase/corporate/pages/vacation_schedule.aspx

One open point is that it seems that Shushan Purim (which is one day after Purim and is typically not observed in Tel Aviv, but e.g. Jerusalem) will be observed from 2022 on, but wasn't observed before. I'm waiting for clarification back from TASE on this point.

Nevertheless, the current version of the code is consistent with what is currently listed on the website at least for 2020 - 2023. I also briefly manually checked a few years before, but no unit tests are there for these yet.

On a technical note, I've re-used Pandas `Easter` class -- which is a subclass of `DateOffset` -- for all holidays since they likewise fall on different Gregorian dates every year. To be able to do that, I had to import a few (internal) Cython functions and classes. A pure Python or even Cython re-implementation seemed to risky in terms of likely forgetting to handle this or that corner-case. We just have to be aware of these imports since they may introduce some more coupling on the specific Pandas version used.

Let me know your thoughts...

### DIFF
--- a/etc/requirements.in
+++ b/etc/requirements.in
@@ -4,3 +4,4 @@ python-dateutil
 pytz
 six
 toolz
+pyluach
--- a/etc/requirements.in
+++ b/etc/requirements.in
@@ -1,7 +1,7 @@
 numpy
 pandas
+pyluach
 python-dateutil
 pytz
 six
 toolz
-pyluach
--- a/etc/requirements_locked.txt
+++ b/etc/requirements_locked.txt
@@ -22,6 +22,7 @@ py-cpuinfo==7.0.0         # via pytest-benchmark
 py==1.9.0                 # via pytest, pytest-forked
 pycodestyle==2.6.0        # via flake8
 pyflakes==2.2.0           # via flake8
+pyluach==1.2.1            # via -r etc/requirements.in
 pyparsing==2.4.7          # via packaging
 pytest-benchmark==3.2.3   # via -r etc/requirements_dev.in
 pytest-forked==1.3.0      # via pytest-xdist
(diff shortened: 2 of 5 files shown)


jenskeiner:
Build is failing because of imports that don't work on older Pandas versions. Not sure if code can be adjusted or of we should generally require a newer Pandas version. Please advise.

jenskeiner:
Semi-fixed the imports with some try...except blocks, but this is very ugly. Also, still fails on Python 2.7.

jenskeiner:
Ah, finally, all checks are passed. I had to use `weekday()` instead of `isoweekday()` on `pyluach` dates since only the former is available when running under Python 2.x with and older `pyluach` version.

jenskeiner:
I got clarification on Shushan Purim from TASE. It is a regular trading day apart from the fact that nothing $-related can be done due to Bank of Israel being closed (I suppose they observe Shushan Purim as headquarter located in Jerusalem). Anyhow, for some reason they added it to the 2022 holidays although it falls on a Friday anyway.

So I've removed Shushan Purim from the list.

gerrymanoim:
Hey @jenskeiner - thanks for doing this all! I'm hoping to clean up all the py2/old pandas compat stuff quite soon. 

At a high level, this looks good to me. I'm generally wary of dependencies, but pyluach looks like it doesn't depend on anything else. 

I want to take a deeper look in the next couple days, but I think this should be good as is. We'll do a big cleanup when everything migrates to py3/pandas 1.

jenskeiner:
Thanks @gerrymanoim. Added 2021 election day now as well.

gerrymanoim (approved):

